PR #1, opened 2019-10-14
Theme setup

Azein:
Added default theme on top of @oyabun existing specs
Reworked a bit the utils with intention to simplify the API.
Stories now live together with components

To be done still - check our application-level import opportunities.


src/media.ts
```diff
@@ -0,0 +1,8 @@
+export const devices = {
```
VLegakis: Should @oyabun confirm these breakpoints?

src/theme/utils.ts
```diff
@@ -0,0 +1,43 @@
+import math from "polished/lib/math/math"
+import { mergeDeepRight, path, pathOr } from "ramda"
+
+export type ThemeAtom = string | number | { [key: string]: string | number | ThemeAtom }
```
VLegakis: Why does this Theme need to opt between all these types?
Wouldn't just a (simple) Theme interface do fine?

builat: @VLegakis coz, there are could be sublevels. Thishow recursive typing implemented here

Azein: Theme could be constructed on application level, with keys/values we have no idea about right now. So, we use a wider type, also setting some constraints.

src/media.ts
Azein: He did.

PR #3, opened 2019-10-16
Button component

VLegakis:
Button component. Demoed here:
http://arachovis.com:6006


package.json
```diff
@@ -24,7 +24,8 @@
     "ramda": "^0.26.1",
     "react": "^16.10.2",
     "react-dom": "^16.10.2",
-    "styled-components": "^4.4.0"
+    "styled-components": "^4.4.0",
+    "rmwc": "^5.7"
```
Azein: Let's install on per-component basis. It's also recommended in rmwc docs.

src/components/button/styled.tsx
```diff
@@ -0,0 +1,21 @@
+import React from "react"
+import styled, { css } from "styled-components"
+import { Button } from "@rmwc/button"
+import { getColor } from "../../theme/utils"
+
+export const StyledButton = styled(({ ...otherProps }) => <Button {...otherProps} />)`
```
Azein: Have you tried just styled(Button)? I have a feeling they improved that.

package.json
VLegakis: Ok. I had it like that in the beginning but then I remember we said that webpack will selectively include what is required only. 
I'll put it back on a separate per component basis...

src/components/button/styled.tsx
VLegakis: But I have conditional CSS bellow based on the props. Would it work as you say? (it doesn't seem so from a fast test I did)

src/components/button/README.md
```diff
@@ -0,0 +1,5 @@
+## Button component
+
+```typescript
+const example: string = "My Button"
```
Azein: As I understood, this was just an example. I suggest to put Props type here to have a quick idea about the types.

src/components/button/styled.tsx
```diff
@@ -0,0 +1,41 @@
...
+  ${props =>
+    props.isLoading
+      ? css`
+          background: ${getColor(["green", "malachite"])};
+          border-radius: 3px;
+          border-color: ${getColor(["green", "malachite"])};
+          text-color:
```
Azein: Why is it empty?

src/components/button/README.md
builat: @VLegakis you can try [this](https://github.com/netdata/netdata-ui/blob/1426-modal/src/components/sidebar/README.md) as example

VLegakis:
Waiting for 
https://github.com/netdata/cloud-frontend/issues/1471 for icon buttons

src/components/button/styled.tsx
VLegakis: By mistake! Fixed

src/components/button/button.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { text, boolean } from "@storybook/addon-knobs"
+import { readmeCleanup } from "../../../utils/readme"
+import { Button } from "."
+
+// @ts-ignore
+import readme from "./README.md"
+import { ButtonType } from "./button"
+import { Icon } from "../icon"
+
+const mockStory = storiesOf("Button", module)
```
builat: why `mockStory` it is't highly important, but anyway

src/components/button/button.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+  jest: ["button.test.tsx"],
+}
+
+mockStory.add(
+  "Button Default",
+  () => {
+    return (
+      <Button
+        disabled={boolean("Disabled", false)}
+        type={ButtonType.default}
```
builat: May be it will be ok, if we set `type` default prop?

src/components/button/styled.tsx
```diff
@@ -0,0 +1,154 @@
...
+import { CircularProgress } from "@rmwc/circular-progress"
+// import "@rmwc/circular-progress/circular-progress.css"
+import { getColor } from "../../theme/utils"
+import { ButtonType } from "./button"
+
+export const StyledButton = styled(({ label, ...otherProps }) => (
+  <Button label={label} {...otherProps} />
+))`
+  ${props => {
+    if (props.disabled) {
```
builat: You can use [propOrElse](https://github.com/netdata/netdata-ui/blob/master/src/components/icon/icon.stories.tsx#L46-L59)

src/components/button/styled.tsx
```diff
@@ -0,0 +1,154 @@
...
+    }
+  }};
+`
+
+export const StyledCircularProgress = styled(({ ...otherProps }) => (
+  <CircularProgress {...otherProps} />
+))`
+  ${() => `
+
+  `};
```
builat: why we need this empty function?

src/components/button/button.stories.tsx
VLegakis: Ooops missed that...
Corrected done.

src/components/button/button.stories.tsx
VLegakis: Yes by default the type is `ButtonType.default` so these line there can be omitted. 
I committed the omitted version so its made more clear.

src/components/button/button.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+    />
+  ),
+  subData
+)
+
+mockStory.add(
+  "Button Icon No Fill",
+  () => (
+    <Button
+      icon={<Icon name={text("Icon name", "plus")} />}
```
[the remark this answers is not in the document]
VLegakis: Done.

src/components/button/styled.tsx
```diff
@@ -0,0 +1,154 @@
...
+        flex-flow: row nowrap;
+        align-items: center;
+        .rmwc-icon {
+          margin-right: ${props.label ? "12px" : "0px"};
+          height: 24px;
+        }
+      `
+    }
+    switch (props.type) {
+      case ButtonType.noFill:
```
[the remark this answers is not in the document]
VLegakis: There will be more button types added.

src/components/button/styled.tsx
VLegakis: Removed

src/components/button/styled.tsx
VLegakis: Ok. I can try that but because of the arrow function here let me try this in one next iteration

src/components/button/styled.tsx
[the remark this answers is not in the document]
builat: may be it will be good idea to use to Map with fallback? to not create everytime switch and not to increase complexity of switching. Coz every new option will be checket every time in worst case. But in Map with fallback complexity is always O(1)

src/components/button/button.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+  readme: {
+    sidebar: readmeCleanup(readme),
+  },
+  jest: ["button.test.tsx"],
+}
+
+buttonStory.add(
+  "Button Default",
+  () => {
+    return (
```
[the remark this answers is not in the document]
VLegakis: Fixed! Thanks @jacekkolasa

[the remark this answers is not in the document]
VLegakis: Fixed! Thanks @jacekkolasa !

src/components/button/styled.tsx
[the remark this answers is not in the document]
VLegakis: Done! Thanks @builat !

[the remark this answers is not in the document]
builat: But where is `Map` based implementation?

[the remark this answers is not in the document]
VLegakis: Its here!: https://github.com/netdata/netdata-ui/pull/3/files#diff-685bc1de267168e93c724e89256ee18d
Perhaps you see the old version in the comments? Refresh?

src/components/button/styled.tsx
```diff
@@ -0,0 +1,266 @@
...
+
+export const StyledButton = styled(({ label, icon, ...otherProps }) => (
+  <Button label={label} icon={icon} {...otherProps} />
+))`
+  ${props => {
+    if (props.disabled) {
+      const template = buttonTemplates.get("disabled")
+      return template ? template(props.label, props.icon, props.isLoading) : css``
+    }
+    const template = buttonTemplates.get(props.type)
```
builat: here could not be prop, so fallback to default should be implemented. Better to create function to access data in `Map`

VLegakis: So the next line is:
`return template ? template(props.label, props.icon, props.isLoading) : css```
which returns a "default" which I want it to be empty CSS. Is this what you mean?

src/components/button/styled.tsx
```diff
@@ -0,0 +1,279 @@
...
+          stroke-dasharray: 185.4%, 416%;
+          stroke-dashoffset: -258.33%;
+        }
+      }
+    `
+  }};
+`
+
+/*
+RMWC ThemeProvider vars
```
Azein: Still needed?

VLegakis: So we don't use any RMWC nor MDC CSS for this implementation. I have put here commented as a reference in case we use some RMWC component and want to see the vars that their CSS is using for styling... We can remove if we don't go with RMWC from now on.
As discussed the experience is rather frustrating. 
Just to mention that I regard it would be the same with any framework.
Lets see if we can go with our own implementations and the effort required there.

PR #4, opened 2019-10-17
Sidebar component

builat:
* Layout with sidebar
* Sidebar based on portals


src/components/sidebar/portal-sidepan.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+  })
+  it(" * should render on the right side", () => {
+    const { queryByText } = testWrapper(
+      PortalSidebar,
+      { right: true, children: "test text" },
+      DefaultTheme,
+      {}
+    )
+    const result = queryByText("test text")
+    expect(result && result.textContent).not.toBeNull()
```
Azein: But it doesn't check it actually rendered right, only that it's just rendered.

src/components/sidebar/portaled-sidebar.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { FC } from "react"
+import { Portal } from "react-portal"
+import { PortalSidebox } from "./styled"
+
+export const PortalSidebar: FC<SidebarProps> = ({ children, right = false }: SidebarProps) => (
+  <Portal>
+    <PortalSidebox shadowXOffset={right ? -2 : 2} side={right ? "right" : "left"}>
```
Azein: The prop name is kinda hard to understand. What is it for?

src/components/sidebar/portal-sidepan.test.tsx
builat: Correct I'll fix it

src/components/sidebar/portaled-sidebar.tsx
builat: It is about dropping shadow

src/components/sidebar/portaled-sidebar.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { FC } from "react"
+import { Portal } from "react-portal"
+import { PortalSidebox } from "./styled"
+
+export const PortalSidebar: FC<SidebarProps> = ({ children, right = false }: SidebarProps) => (
+  <Portal>
```
Azein: How can we close the portal? Do we have any opportunities like overlay click, escape press?

src/components/sidebar/portaled-sidebar.tsx
Azein: And why is it 2 or -2 numeric?

builat: This exact values from figma

src/components/sidebar/portaled-sidebar.tsx
builat: @Azein implemented

src/components/sidebar/portaled-sidebar.tsx
builat: @Azein  resolved

src/components/sidebar/types.d.ts
```diff
@@ -0,0 +1,15 @@
...
+  right?: boolean
+  children?: NodeT
+}
+
+interface PortalSidebarProps<T = any> {
+  isOpen?: boolean
+  closeOnEsc?: boolean
+  onClose?: (args?: T) => void
+  right?: boolean
+  children?: NodeT
```
VLegakis: Would a `disabled` optional property make sense here so we can disable it?

src/components/sidebar/types.d.ts
```diff
@@ -0,0 +1,15 @@
...
+interface SidebarProps {
+  info?: NodeT
+  right?: boolean
+  children?: NodeT
+}
+
+interface PortalSidebarProps<T = any> {
+  isOpen?: boolean
+  closeOnEsc?: boolean
+  onClose?: (args?: T) => void
```
VLegakis: Wouldn't it be better to have a `onToggle` with a `isOpen` boolean param so we have a handler also when it opens?

builat: @VLegakis  `isOpen` is state, `onToggle` as far it is sidebar all kind of controls for it is in the parent component

src/components/sidebar/types.d.ts
builat: is `isOpen` setted to false is eq to your proposal?

src/components/sidebar/portaled-sidebar.tsx
```diff
@@ -0,0 +1,41 @@
...
+      document.removeEventListener("keydown", (event: KeyboardEvent) => {
+        if (event.keyCode === ESCAPE_KEY && closeOnEsc) {
+          setOpen(false)
+          onClose()
+        }
+      })
+    }
+  }, [isOpen])
+
+  return isOpenState ? (
```
Azein: Does it mean we want to have an uncontrolled component? If so, why we still provide isOpen, and why it defaults to true? Can we just assume the panel will be controlled from above, and remove the local state, defaulting etc from inside? 
If we still want this functionality, it's easier to write a wrapper.

src/components/sidebar/sidebar.stories.tsx
```diff
@@ -0,0 +1,117 @@
...
+)
+
+const SidebarContent = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  background-color: #fff;
+  height: 100%;
+  width: 100%;
+  color: black;
```
Azein: There is no black in the theme.

src/components/sidebar/portaled-sidebar.tsx
```diff
@@ -0,0 +1,41 @@
...
+        setOpen(false)
+        onClose()
+      }
+    })
+    return () => {
+      document.removeEventListener("keydown", (event: KeyboardEvent) => {
+        if (event.keyCode === ESCAPE_KEY && closeOnEsc) {
+          setOpen(false)
+          onClose()
+        }
```
jacekkolasa: i think that will not remove event listener properly, you think to pass the exact same function

builat: fixed

src/components/sidebar/portaled-sidebar.tsx
builat: fixed after discuss

src/components/sidebar/portaled-sidebar.tsx
```diff
@@ -0,0 +1,41 @@
...
+      })
+    }
+  }, [isOpen])
```
jacekkolasa: IMO `closeOnEsc` should be added here, in case that prop changes, so keydown listener is refreshed. btw. linter should shows me that exact error (`react-hooks/exhaustive-deps`)

builat: done

src/components/sidebar/types.d.ts
builat: `isOpen` removed. As far as all controls will be in the parent component.

src/components/sidebar/types.d.ts
builat: isOpen removed. As far as all controls will be in the parent component.

src/components/sidebar/sidebar.tsx
```diff
@@ -0,0 +1,9 @@
+import React, { FC } from "react"
+import { SidebarBox, ComponentBox, InfoBox } from "./styled"
+
+export const Sidebar: FC<SidebarProps> = ({ info, children, right = false }: SidebarProps) => (
+  <ComponentBox side={right ? "row-reverse" : "row"}>
```
Azein: I suggest to pass the `right` prop into styled and pick the style there. "row-reverse" and other css syntax shouldn't appear on component level.

builat: @Azein done

src/components/sidebar/types.d.ts
```diff
@@ -0,0 +1,26 @@
+type NodeT = import("react").ReactNode
```
VLegakis: Why do we use this NodeT here instead of a `ReactNode` ?

builat: Just to write less

VLegakis: Just a minute we introduce a new type just because it has a smaller and more handy name? I think this is not correct. I would rename a lot of names with names I prefer but this would make my code much less readable. I can create e.g a CompT for ReactComponent etc because its easier to write but this would make my code not very easy to read for others I think

VLegakis:
Also, The `opened - text should disappear after you press Esc` text is not centered in the left panel but rather hidden from the right panel (attached screenshot) along with some strange gap in the bottom of the right panel (not sure if this behavior expected)

Screenshot 2019-10-21 at 13 34 16 copy

src/components/sidebar/types.d.ts
VLegakis: So, if `isOpen` is removed and the controls reside on the parent, what does `onClose` do? Can you elaborate a bit on this?

builat: side effect wich will be fired on close

PR #5, opened 2019-10-17
Checkbox

Azein:
Deployed for review:
http://arachovis.com:6007


src/components/checkbox/README.md
```diff
@@ -0,0 +1,18 @@
...
+```typescript
+interface CheckboxProps {
+  checked: boolean
+  onChange: (e: React.ChangeEvent<HTMLInputElement>) => void
+  id?: string
+  label?: string
+  labelPosition?: "left" | "right"
+  ref?: React.MutableRefObject<HTMLInputElement>
+  className?: string
+}
```
VLegakis: The props should include also the `disabled` property for disabling the control (and the relevant CSS handling of this).

src/components/checkbox/README.md
```diff
@@ -0,0 +1,18 @@
...
+### Props:
+
+```typescript
+interface CheckboxProps {
+  checked: boolean
+  onChange: (e: React.ChangeEvent<HTMLInputElement>) => void
+  id?: string
+  label?: string
+  labelPosition?: "left" | "right"
+  ref?: React.MutableRefObject<HTMLInputElement>
```
VLegakis: Also, should we expose the `ref` and `className`? I mean we can expose all the underling properties of course but maybe we should agree on which are useful for us and avoid the rest. We can agree on a common set of required props like for example:
`id` and `className`

jacekkolasa:
LGTM, but i'm not sure if this is proper vertical centering?
Chrome:
image

Firefox:
image

src/components/checkbox/README.md
Azein: `ref` could be exposed so the outer code can manipulate it (with useRef etc). `className` is exposed so that the Checkbox will be wrappable in styled()

src/components/checkbox/README.md
Azein: I agree with you, I just didn't see the disabled state in Figma, will ask @oyabun

src/components/checkbox/README.md
VLegakis: Ok. So @builat and @Azein lets agree that all our components will have:
`id`, `ref` and `className`

src/components/checkbox/README.md
Azein: Done

Azein:
http://arachovis.com:6007/ link to review

src/components/checkbox/README.md
VLegakis: Nice! Thanks! Only thing to discuss would be if we need to have the disabled state as an explicit story or as a knob control to avoid too many stories. TBD

Azein:
@jacekkolasa 

I added some explicit centering for now, anyway might be adjusted after typography addition, thanks for cross-browser testing btw =)

src/components/checkbox/README.md
Azein: Using knob now

PR #13, opened 2019-10-24
CSS reset utility

Azein:
Set of reset rules, according to https://dev.to/hankchizljaw/a-modern-css-reset-6p3
Intended to be used across projects relying on UI Kit, we may also expand this ruleset.

### DIFF
--- a/src/styles-reset.ts
+++ b/src/styles-reset.ts
@@ -1,0 +1,82 @@
+import { css } from "styled-components"
+
+export const stylesReset = css`
+  /* Box sizing rules */
+  *,
+  *::before,
+  *::after {
+    box-sizing: border-box;
+  }
+
+  /* Remove default padding */
+  ul[class],
+  ol[class] {
+    padding: 0;
+  }
+
+  /* Remove default margin */
+  body,
+  h1,
+  h2,
+  h3,
+  h4,
+  p,
+  ul[class],
+  ol[class],
+  li,
+  figure,
+  figcaption,
+  blockquote,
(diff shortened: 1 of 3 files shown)


src/styles-reset.ts
```diff
@@ -0,0 +1,82 @@
...
+    margin: 0;
+  }
+
+  /* Set core body defaults */
+  body {
+    min-height: 100vh;
+    scroll-behavior: smooth;
+    text-rendering: optimizeSpeed;
+    line-height: 1.5;
+  }
```
jacekkolasa: i agree with having the styles-reset, but shouldn't `body` styles be placed in the specific app?

Azein: They should be placed in the specific app, it's just a module to import and use, if you like. For instance, here in storybook app, I use them through GlobalStyles styled-components API. Other people may or may not use it, it's not "built-in" in the ui-kit, just like media queries etc.

jacekkolasa: I understand, i had a feeling that `min-height: 100vh` is too specific for the individual page, and it can be problematic to treat it as an "universal solution" for us. The common style reset is very helpful and it would be not good to create separate resets in our apps, but on the other hand, if needed, we can override problematic style if it will be needed.

Azein: We can even remove the problematic style, as this dependency is meant to be used by us on the first place =)

Azein:
Closing this, as application-side approach was implemented long time ago.

PR #14, opened 2019-10-25
Unified conditional CSS experiment

VLegakis:


src/theme/default/colors.ts
```diff
@@ -1,6 +1,7 @@
 const rawColors: RawColorsT = {
   transparent: {
-    transparent: "transparent",
+    fullyTransparent: "rgba(0, 0, 0, 0.0)",
```
Azein: transparent seems not to be a "color".

PR #15, opened 2019-10-25
1425 TextInput component

Azein:
The idea is to keep TextInput as dumb as possible, and able to consume values from wrappers/hooks. However, we might need to expose some convenience hooks/wrappers by default, as I did with useTouchedState. Something like useInputValue might be useful also. Let's think about it.


src/hooks/use-touched-state.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+type BlurHandler = (e: FocusEvent) => void
+
+type UseTouchedStateHookValues = [
+  boolean,
+  BlurHandler,
+  React.Dispatch<React.SetStateAction<boolean>>
+]
+
+export const useTouchedState = (
```
Azein: I think I will move it under the input directory, seems to be not generic enough.

src/theme/utils.ts
```diff
@@ -26,7 +26,7 @@ export const calcSize = (expr?: string) => (props: WrappedTheme) => {
 
 export const getColor = (colorPath: string[]) => getOrElse(["colors", ...colorPath], "#fff")
 
-export const getSizeBy = (multiplier: number) => (props: WrappedTheme) => {
+export const getSizeBy = (multiplier: number = 1) => (props: WrappedTheme) => {
   const size = (getSizeUnit(props) || 0) * multiplier
```
builat: 0 is not the best fallback value for sizes. May be will be better to implement fallback in **getSizeUnit** function?
```typescript
  const size = (getSizeUnit(props) || 1) * multiplier
```

Azein:
Deployed on http://arachovis.com:6007 for design review

src/components/input/input.stories.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { text, number, boolean } from "@storybook/addon-knobs"
+import { useCallback } from "@storybook/addons"
+import { getColor } from "../../theme/utils"
+import { readmeCleanup } from "../../../utils/readme"
+import { TextInput, useInputValue, useTouchedState } from "."
+import { Icon } from "../icon"
+// @ts-ignore
+import readme from "./README.md"
+
+const inputStory = storiesOf("TextInput", module)
```
VLegakis: I have organised the storybook components into categories. (Components, Utils and Layout for the moments). So maybe this line would be good to be set as: `storiesOf("COMPONENTS|Controls/TextInput", module)`

Azein: I suggest to merge as is (master doesn't contain the change you made yet), and apply the structure later as patch.

PR #16, opened 2019-10-29
Typography

VLegakis:
Demo
http://arachovis.com:6006/
Mock:
https://www.figma.com/file/28Xv1J95zLbPowJujLuArZ/ND-Design-System-0.4?node-id=8%3A224


VLegakis:
> Implementation looks okay, the retina fonts is for @oyabun to decide. There are two things to pay attention:
> 
> 1. Is the Typography included into the ts-bundle? Maybe I just missed the index tho.
> 2. I suggest to have 2 exports from each Typography element - the styles in form of `css` mixin from styled-components, and the element itself (like already done). This might be useful for some controls styling which do not directly use the text element, but might need the styles (like placeholder etc).

Done with mixins. Also, added some bigger sizes which I was just hiding before although they still existed.
Let me know if it looks ok.

Issue #25, opened 2019-11-06
Button componet errors


builat:
* Component cannot be overrided. 
```typescript
// For example
// this does not works
const MyBtn = styled(Button)`
  font-size: 200px;
`
```
* If I will render button without additional props I will have:
![image](https://user-images.githubusercontent.com/3184538/68307435-d1737a80-00dd-11ea-9234-6edbdad16992.png)

Also, not a bug, but not usefull for me. I need to import `enum` of button types instead of writing it as simple string.

VLegakis:
Thanks Dmitry. Let me think a bot about this. So you say you want to override the styles of buttons? This may be possible but isn't it against the reason we create the UIKit so we have consistent styling? Maybe for your case I could create another button type which has the styles you want?

builat:
@VLegakis I just want to override some part of styles not a lot of them. So you will never can create all the variations of button, so more rational to give user posibility of overriding.

Also if we talking about "SKIP" `borderless` does not works either.

builat:
Closed via PR: #27

PR #27, opened 2019-11-07
Closes issue #25

builat:
* Closes issue #25 
* Resolves button styled override
* Fixes some perf issues
* Fixes button types interface

### DIFF
--- a/src/components/button/README.md
+++ b/src/components/button/README.md
@@ -19,11 +19,8 @@ interface Props {
 ## Button types
 
 ```typescript
-export enum ButtonType {
-  default = "default",
-  noFill = "noFill",
-  borderless = "borderless",
-}
+  type ButtonType = "default" | "noFill" | "borderless",
+
 ```
 
 ## Typical usage
@@ -44,16 +41,10 @@ const myButton: Button = (
 
 ### Icon button
 
-Please note `type={ButtonTypes.default}` is optional (can be omitted):
+Please note `"default"` is optional (can be omitted):
 
 ```typescript
 const myButton: Button = (
-  <Button
-    disabled={true}
-    type={ButtonTypes.default}
-    icon={"plus"}
-    onClick={this.myHandler}
-    label="Press me"
-  />
+  <Button disabled={true} type="default" icon={"plus"} onClick={this.myHandler} label="Press me" />
 )
 ```
--- a/src/components/button/button.tsx
+++ b/src/components/button/button.tsx
@@ -1,12 +1,8 @@
 import React, { SyntheticEvent, FC } from "react"
 import { StyledButton, StyledCircularProgress } from "./styled"
-import { Icon } from "../icon"
+import { DEFAULT } from "./constants"
 
-export enum ButtonType {
-  default = "default",
-  noFill = "noFill",
-  borderless = "borderless",
-}
+type ButtonType = "default" | "noFill" | "borderless"
 
 export interface ButtonProps {
   id?: string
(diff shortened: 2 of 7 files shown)


src/components/button/README.md
```diff
@@ -19,11 +19,8 @@ interface Props {
 ## Button types
 
 ```typescript
-export enum ButtonType {
-  default = "default",
-  noFill = "noFill",
-  borderless = "borderless",
-}
+  type ButtonType = "default" | "noFill" | "borderless",
```
VLegakis: Nope. Why is this required to change?

src/components/button/styled.tsx
```diff
@@ -176,7 +131,7 @@ export const StyledButton = styled(({ label, icon, ...otherProps }) => (
     opacity: ${props.disabled ? 0.6 : 1.0};
     cursor: ${props.disabled ? "not-allowed" : "pointer"};
     background-color: ${buttonProps("buttonColorNormal", props)};
-    border-color: ${getColor(["green", "greenHaze"])}
+    border-color: ${getGreenHaze}
```
Azein: Easier to change with just getColor, let's not write a special named function to get a certain color.

src/components/button/button.tsx
```diff
@@ -25,17 +21,18 @@ export const Button: FC<ButtonProps> = ({
   onClick = () => {},
   icon = null,
   isLoading = false,
-  type = ButtonType.default,
+  type = DEFAULT,
```
VLegakis: Why take this from a string constants file instead of the original enum?

src/components/button/button.tsx
```diff
@@ -25,17 +21,18 @@ export const Button: FC<ButtonProps> = ({
...
   disabled = false,
+  ...rest
```
Azein: let's use ...props

src/components/button/README.md
```diff
@@ -44,16 +41,10 @@ const myButton: Button = (
...
 ```typescript
 const myButton: Button = (
-  <Button
-    disabled={true}
-    type={ButtonTypes.default}
-    icon={"plus"}
-    onClick={this.myHandler}
-    label="Press me"
-  />
+  <Button disabled={true} type="default" icon={"plus"} onClick={this.myHandler} label="Press me" />
```
Azein: I suggest to use icon the same way as in <TextInput>, thus passing the component, not a string.

src/components/button/button.tsx
```diff
@@ -25,17 +21,18 @@ export const Button: FC<ButtonProps> = ({
   onClick = () => {},
```
Azein: Let's mark returning type as void, () => {} seems to be a bit misleading

src/components/button/styled.tsx
```diff
@@ -2,127 +2,82 @@ import React from "react"
...
+const getGreenMalachite = getColor(["green", "malachite"])
 
 const buttonPropsMap = new Map<string, (props: ButtonProps) => any>([
-  [
-    "divWidth",
-    props => {
-      return props.label ? "134px" : "46px"
-    },
-  ],
+  ["divWidth", ({ label }) => (label ? "134px" : "46px")],
```
Azein: px should be removed in favour of getSizeBy / other means of getting size from theme-defined constant.

src/components/button/README.md
Azein: As I understood, the inconvenience is that you have to import enum from the component to set the type other than default. Correct me if I'm wrong. If so, I'm 100% for this change, you should not be obliged to import enum types from UI component module to be able to set some customization.

builat: Why should I do
```typescript
import {Button, ButtonTypes } from '@netdata/netdat-ui'
function Component() {
  return(<>
    <Button label="any" type={ButtonType.noFill} .../>
  </>)
}
```
Instead of just "noFill". A lot of redundant extra complexity

src/components/button/button.tsx
builat: done

src/components/button/button.tsx
jacekkolasa: @Azein i think it is void, it's not type definition, it's the default value here

PR #35, opened 2019-11-15
autoFocus, ref prop name changed, lazy ref example

Azein:


src/components/input/input.tsx
```diff
@@ -16,17 +16,20 @@ import {
...
 
+type CallBackRef = (input: any) => void
+
 export interface InputProps {
   onChange: (e: ChangeEvent<HTMLInputElement>) => void
   onFocus?: (e: FocusEvent) => void
   onBlur?: (e: FocusEvent) => void
   value: string
-  ref?: MutableRefObject<HTMLInputElement>
+  inputRef?: MutableRefObject<any> | CallBackRef
```
builat:
```typescript
export interface InputProps<T = any> {
  ...
  inputRef?: MutableRefObject<T> | CallBackRef
}
  ...
```

builat: May be this could be better for strict types?

PR #36, opened 2019-11-18
added optional onclick to icon

Azein:


VLegakis:
This is not required @Azein 
It can be handled like e.g:
```
<Button type="borderless" icon="trashcan" onClick={() => handleDelete({ email, name })} />
```

PR #37, opened 2019-11-18
Typography theming

VLegakis:
Some color updates in the UIKit

### DIFF
--- a/src/components/typography/typography.stories.tsx
+++ b/src/components/typography/typography.stories.tsx
@@ -28,7 +28,7 @@ const subData = {
 
 const headers = (
   <div>
-    <H0>H1: Title Enourmous. The five boxing wizards jump quickly.</H0>
+    <H0>H0: Title Enourmous. The five boxing wizards jump quickly.</H0>
     <H1>H1: Title Huge. The five boxing wizards jump quickly.</H1>
     <H2>H2: Title Bigger. The five boxing wizards jump quickly.</H2>
     <H3>H3: Title Big. The five boxing wizards jump quickly.</H3>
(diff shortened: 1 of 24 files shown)


src/theme/default/colors.ts
```diff
@@ -42,10 +42,10 @@ const rawColors: RawColorsT = {
 
 const appColors: AppColorsT = {
   borderColor: rawColors.gray.bombay,
-  controlFocused: rawColors.gray.limedSpruce,
+  controlFocused: rawColors.gray.darkMud,
   error: rawColors.red.redOrange,
   success: rawColors.green.greenHaze,
-  text: rawColors.gray.limedSpruce,
+  text: rawColors.gray.darkMud,
```
Azein: Was the previous value wrong?

VLegakis: All the latest mocks have this color for text (as well as `HeadPanel` etc...)

VLegakis: Resolving after discussion with @oyabun concluding in using 335414A

src/components/typography/typography.stories.tsx
```diff
@@ -28,7 +28,7 @@ const subData = {
 
 const headers = (
   <div>
-    <H0>H1: Title Enourmous. The five boxing wizards jump quickly.</H0>
+    <H0>H0: Title Enourmous. The five boxing wizards jump quickly.</H0>
```
Azein: There is no h0 DOM element. Name should be changed I think.

VLegakis: So is the case for H1, H2 etc they don't exist. 
In any case propose a name and I can rename it.
Also, this is not related to this PR.

PR #42, opened 2019-11-28
Table component

builat:
* implementation was remastered after all


src/components/icon/assets/arrow_down.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="16" height="16" viewBox="0 0 16 16"  xmlns="http://www.w3.org/2000/svg">
```
VLegakis: Maybe there should also be an arrow up for toggling sorting order. But ok we can put these details later

src/components/table/styled.tsx
```diff
@@ -0,0 +1,53 @@
+import styled from "styled-components"
+import { getSizeBy, getColor } from "../../theme/utils"
+import { Icon } from "../icon"
+import { Button } from "../button"
+
+export const StyledTable = styled.table`
+  width: 100%;
```
VLegakis: I think the table doesn't resize its width once the parent container resizes.

src/components/table/table.tsx
```diff
@@ -0,0 +1,82 @@
+import React, { useEffect } from "react"
+import { useTable, useSortBy, useRowSelect, ColumnInstance, Row } from "react-table"
+import { UserHeader } from "./components/UserHeader"
+
+import { StyledTable, StyledThead, StyledRow, RowBox } from "./styled"
+
+interface TablePropsT<T, RT = any> {
```
VLegakis: This is an Interface (no T required in the name)

src/components/table/table.tsx
```diff
@@ -0,0 +1,82 @@
...
+  columns: RT
+  data: T[]
+  sortedBy?: string[]
+}
+
+function RTable<T extends object>({
+  columns,
+  data,
+  sortedBy = [],
+  selectedItemsClb,
```
VLegakis: What does Clb stand for?

builat: callbcack

src/components/icon/assets/arrow_down.svg
builat: May be, but it is all I have at moment.

image

builat:
I can add example of infinit scrolling, but it probably not the part of the table, but part of `row` component inside

src/components/icon/assets/arrow_down.svg
VLegakis: Ok. so if sorting order changes the arrow remains the same right? We can fix that in some next iteration

PR #44, opened 2019-12-04
input reset and custom meta display

Azein:
Includes following features:

- Better imperative handling for useInputValue hook
- Ability to use arbitrary function for showing validation feedback, apart from ‘all’ and ‘positiveFirst’.

### DIFF
--- a/src/components/input/use-input-value.tsx
+++ b/src/components/input/use-input-value.tsx
@@ -1,10 +1,13 @@
-import { useState, useCallback } from "react"
+import { useState, useCallback, useMemo } from "react"
 import { ChangeEventHandler, ReactInputChangeEvent } from "./types"
 
 type InputValue = string
 type MaxCharsIndicator = string
 type IsDirty = boolean
 type SetIsDirty = React.Dispatch<React.SetStateAction<boolean>>
+type SetValue = React.Dispatch<React.SetStateAction<string>>
+type ResetValue = (value?: string) => void
+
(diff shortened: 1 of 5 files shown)


src/components/input/use-input-value.tsx
```diff
@@ -38,7 +51,21 @@ export const useInputValue: UseInputValue = ({ value = "", onChange, maxChars })
     [isDirty, maxChars, onChange]
   )
 
-  const maxCharsIndicator = maxChars ? `${inputValue.length}/${maxChars}` : ""
+  const maxCharsIndicator = useMemo(() => (maxChars ? `${inputValue.length}/${maxChars}` : ""), [
+    maxChars,
+    inputValue,
+  ])
```
jacekkolasa: @Azein not sure if that simple operation needs `useMemo` (maybe `useMemo` perf cost is bigger)

Azein: Almost no single operation needs useMemo to not do the operation, unless it's some very big iteration. The purpose is not to trigger re-render of the tree that will use the maxCharsIndicator value from this hook, which may have considerable cost (or may not, but this is the UI Kit level, so we probably should care).

We are still not doing this in all cases, but we should start to care about memo-ing the stuff we expose, I think.

jacekkolasa: i don't understand, when you return primitives, `useMemo` shouldn't impact rerender, or am i missing something?

PR #47, opened 2019-12-09
Fix icons

### LINKED ISSUE #46: Button icon: broken fill, hardcoded icon size
Fill value for icon has extra `,` symbol, which breaks the css rule.

Also, the icon size of 24px seems to be hardcoded, which makes tricky working with smaller/larger icons. Are there some workarounds, or this is something we should improve?

VLegakis:
This issue fixes #46 
Most of the icons had an issue since they had a hard coded fill  @oyabun


src/components/button/styled.tsx
```diff
@@ -126,18 +126,10 @@ export const StyledButtonWrapper = styled.div<{ label: string }>`
...
-    `
-  }};
-`
-
-export const StyledButton = styled(({ label, icon, ...otherProps }) => (
+export const StyledButton = styled(({ label, icon, danger, ...otherProps }) => (
   <StyledButtonWrapper label={label}>
-    <button type="button" {...otherProps}>
-      {icon ? <StyledIcon name={icon} {...otherProps} /> : null}
+    <button type="button" danger={danger.toString()} {...otherProps}>
```
Azein: Why is .toString() needed?

src/components/button/styled.tsx
```diff
@@ -126,18 +126,10 @@ export const StyledButtonWrapper = styled.div<{ label: string }>`
...
+      {icon ? <Icon className="button-icon" name={icon} /> : null}
```
Azein: What is the specific reason to use classname?

VLegakis: This is how this component is selected in the CSS

src/components/button/styled.tsx
VLegakis: As far as I've seen this was the error hitting since danger is boolean.

src/components/button/styled.tsx
```diff
@@ -150,7 +142,7 @@ export const StyledButton = styled(({ label, icon, ...otherProps }) => (
     &button {
       padding: 0;
     }
-    padding: 8px;
+    padding: ${getSizeBy(1)};
```
jacekkolasa: i don't understand why this is an improvement, is there a possibility that we will be changing all sizes in the application by changing `SIZE_UNIT`? Correct me if i'm wrong, but i think it's very unlikely. If we really need wrapper, perhaps we could at least use the same number, without having to multiply/divide by 8 all the time, when comparing this to designs?

builat: could be just `getSizeBy()`

Azein: It's a consistent practice that allows to stick with defined measurements. Any wrong size in the mock is rounded to agreed measurements, unless it makes no sense (small icons etc). So, even if we won't change the base size unit ever, I believe it's a nice practice to add more consistency to the layouts.

src/components/button/styled.tsx
Azein: I don't understand it still. Why we need the "danger" flag converted to string, what is happening overall? I get that there was an error, but what's the reason we need this logic?

src/components/button/styled.tsx
jacekkolasa: thx for explanation, it wasn't clear to me that you agreed to stick with multiplicities of 8 and that is the purpose of those functions. It just looked a little over-engineered (IMHO), a little complication to trivial "px" value, but of course there's no problem to getting used to it :+1:. Oh, and personally i prefer explicit (1) than a default number, just one less thing to worry about

PR #63, opened 2020-01-22
Block layout / scrollable container

Azein:


Azein:
Introduces alternative layout for the table, scrolling in both directions, custom grouping, and some tested workarounds to apply having in mind certain designs (table row layout override on application level). Readme is updated, however types are pretty rough (as `react-table` types or currently not very reliable and library is being re-written to TS).
Basically, any react-table entity like `column` or `row` instances are counted as any, for internal API library docs have to be consulted with.

PR #67, opened 2020-02-18
159 disable modals background

VLegakis:
This blurs and disables interactions with portal's background content


src/components/icon/assets/node_slave.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
Azein: Are you sure it's not already added? As "node" or something else? I'm almost sure we are using this already, you can cycle through existing ones in storybook.

VLegakis: Let me recheck! Maybe I missed that this already exists

VLegakis: Yeap! It existed. I removed it! Thanks @Azein

VLegakis:
> Current implementation isn't working. Layouting is wrong, as overlay blurs the content of portal, and also prevents any click events (because of pointer-events set to none), which we don't want for the portaled sidebar.

Indeed I had implemented the other way round by mistake! Disabling the portal's content instead of its background! Got confused with the right-left side somehow :( fixed now. Please take a look

PR #74, opened 2020-04-06
Added icons

VLegakis:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/netdata-ui",
-  "version": "0.7.26",
+  "version": "0.7.27",
```
jacekkolasa, line 3: don't you need to add also package-lock.json?

src/components/icon/assets/alarm_c.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
Azein, line 1: `fill="none"` means it has hardcoded fill color. As intended?

VLegakis: You are right! It was the way silly figma exportes it. Removed!

PR #76, opened 2020-04-29
Filtering plugin

Azein:
Looking at the application functionality now, I concluded that we can start with global filtering (that's what we do currently), and if this approach works, we can, if needed, add per-column filtering, with another  `react-table` plugin.

### DIFF
--- a/src/components/table/README.md
+++ b/src/components/table/README.md
@@ -14,16 +14,8 @@ from flattened array, or to render groups some other way.
 
 ```typescript
 interface TableProps<T, RT = any> {
-  groupsOrderSettings?: {
-    groupsOrder: {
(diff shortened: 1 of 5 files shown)


src/components/table/README.md
```diff
@@ -32,14 +24,16 @@ interface TableProps<T, RT = any> {
   autoResetSelectedRows?: boolean
   autoResetSortBy?: boolean
   autoResetGroupBy?: boolean
+  autoResetFilters?: boolean
+  // initializer for table instance state, according to react-table signature
+  initialState?: {
+    sortBy?: [{ id: string; desc: Boolean }]
```
builat, line 30: boolean should be `boolean`

PR #78, opened 2020-05-25
Virtualized table

Azein:
Currently it offers virtualized table as separate component, which probably shouldn't be the case and could be avoided without extra complexity, but this should be checked in real world application. There aren't many points of divergence, maybe only passing a custom row renderer, and unavailable callbackRef, but more could be added with extra needs I can't predict atm.


VLegakis:
Why should virtualised table be a separate component?
As you mention it can be just one component or if you want to play extra safe can't it just be an extension of the existing table we have?
Also, I remember you mentioned there will be some callback  (short of local pagination) implementation decided on a meeting I wasn't in. Can you clarify what happened with that or whether it will make it in this PR?

Azein:
@VLegakis In the end, I'd like to have it integrated into the original component, but there are some points of divergence, so it makes sense to proceed after this component will be tested by integration in the app. Right now I can't tell what exact changes I will make and if it will result in harder "merging in original component" process.

As for callback, it will be added in separate iteration, and probably it will be completely on the application side of things, with this component only as data consumer.

PR #87, opened 2020-06-23
Toggle component

novykh:
Introduced a new toggle component according to  https://www.figma.com/file/f7kwppd2Wt8xYi6FSOR2XK/%F0%9F%93%A6Netdata-Design-Elements

I did a custom/simple implementation without any UI framework, since toggle is just a fancy checkbox.

I added an extra prop for the Label component, in case we want to change the label according to the typography components (by default it's the normal 14px Text). If you feel in won't be used I can remove it before merge, or if it's going to be used, I can make the same change for checkboxes as well.

As a final note, do we care about uncontrolled toggles?

### DIFF
--- a/src/components/toggle/styled.tsx
+++ b/src/components/toggle/styled.tsx
@@ -1,0 +1,86 @@
+import React from "react"
+import styled from "styled-components"
+import { getColor, getSizeUnit } from "../../theme/utils"
+import { controlFocused } from "../../mixins"
+
+export const ToggleContainer = styled.div`
+  display: block;
+  box-sizing: border-box;
+  width: 40px;
+  height: 20px;
+`
+
+export const HiddenToggleInput = styled.input.attrs({
+  type: "checkbox",
+})`
+  display: none;
+`
+
+export const StyledToggle = styled.div<{
+  checked: boolean
+  disabled?: boolean
+  colored?: boolean
+}>`
+  box-sizing: border-box;
+  width: 40px;
+  height: 20px;
+  background: ${getColor(["white", "pure"])};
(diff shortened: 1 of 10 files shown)


src/components/toggle/styled.tsx
```diff
@@ -0,0 +1,86 @@
...
+    content: "";
+    width: 16px;
+    height: 16px;
+    border-radius: 50%;
+    left: 5%;
+    top: 50%;
+    transform: translateY(-50%);
+    transition: left 0.2s ease;
+    background-color: ${({ disabled, colored, checked }) => {
+      if (disabled) return getColor(["gray", "gainsboro"])
```
builat, line 52: If you want to use explicit call of `getColor` you have to provide second call with `theme`

This implementation is about this usage:
```typescript
// implicit way
color: ${getColor(["gray", "gainsboro"])};
// explicit way
color: ${({theme, ...props}) => getColor(["gray", "gainsboro"])({theme})};
```

VLegakis:
> Introduced a new toggle component according to https://www.figma.com/file/f7kwppd2Wt8xYi6FSOR2XK/%F0%9F%93%A6Netdata-Design-Elements
> 
> I did a custom/simple implementation without any UI framework, since toggle is just a fancy checkbox.
> 
> I added an extra prop for the Label component, in case we want to change the label according to the typography components (by default it's the normal 14px Text). If you feel in won't be used I can remove it before merge, or if it's going to be used, I can make the same change for checkboxes as well.
> 
> As a final note, do we care about uncontrolled toggles?

Regarding the last note: we can re-iterate that at some point in the future but for the moment we don't care about uncontrolled components

PR #88, opened 2020-06-25
Button component redesign

novykh:
What to expect for while going through this PR:
- Redesigned buttons.
- Added new flavor prop, in order to deprecate type prop, since type is a valid button prop.
- Added new `small` prop to work for action buttons (meaning only icon buttons).
- Added `loadingLabel` prop in order to be able to show loading text when `isLoading` is true.
- Updated stories, I actually kept just two stories - one for the default behaviour and one a css-overridden button - and added all the knobs available for both of them.
- Added test to reach 100% coverage for the component.

** This is still a WIP while waiting for the loading svg icon (or a css only loading approach) to put as a default value. **


src/components/button/button.test.tsx
```diff
@@ -3,34 +3,242 @@
...
+
+  it(" * should render only icon", () => {
+    const { container } = testWrapper<ButtonWrapperProps>(
+      Button,
+      { type: "hollow", icon: "plus" },
+      DefaultTheme,
+      null
+    )
+    const button = container.firstChild
+    expect(button).toMatchSnapshot()
```
Azein, line 55: Nice work with the tests. Are you sure we can benefit from snapshots?

VLegakis:
This look nice apart from a missing icon I think (upon build @novykh ):
```
ERROR in ./src/components/icon/icons-list.ts
Module not found: Error: Can't resolve './assets/loader.svg' in '/Users/vlegakis/Projects/Netdata/netdata-ui/src/components/icon'
```

PR #89, opened 2020-06-26
Virtualized scroll callbacks

Azein:
Some additional handlers in addition to `react-window` render callback, so we can receive the already flattened/ordered rows on app level.

### DIFF
--- a/src/components/table/virtualized-table.tsx
+++ b/src/components/table/virtualized-table.tsx
@@ -1,5 +1,4 @@
-import React, { useEffect, useMemo, useCallback, useState } from "react"
-import { useDebounce } from "react-use"
+import React, { useEffect, useMemo, useCallback } from "react"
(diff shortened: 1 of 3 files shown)


src/components/table/virtualized-table.tsx
```diff
@@ -193,7 +205,7 @@ export function VirtualizedTable<T extends object>({
         orderedRows={orderedRows}
         innerRef={innerRef}
         outerRef={outerRef}
-        onItemsRendered={onItemsRendered}
+        onItemsRendered={itemsRenderHandler}
```
jacekkolasa, line 208: i would name it `handleItemsRendered` (replacing `on` with `handle`), but we don't have any convention so i'm ok with that. We can someday discuss if we want to follow rules like here: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-handler-names.md

PR #91, opened 2020-06-29
Use select logic with react table

novykh:
- Fixed selection issues.
- Introduced `itemIsDisabled` function for easy tracking of rows that are disabled based on a row property.
- Introduced `toggleSelectedItemClb` that get triggered when a single row is getting selected/deselected.
- Added tests.
- Fixed an issue with row rendering for grouping state.


src/components/table/table.tsx
```diff
@@ -113,16 +128,29 @@ export function Table<T extends object>({
...
     },
     ...reactTableHooks
   )
 
   useEffect(() => {
-    if (selectedItemsClb) {
-      selectedItemsClb(selectedFlatRows.map((r: Row<T>) => r.original))
+    if ((selectedFlatRows.length === 0 || isAllRowsSelected) && selectedItemsClb) {
+      const isGrouped = groupBy.length > 0
+      selectedItemsClb(
```
Azein, line 140: I would suggest to not put a lengthy function code inside the call itself. Maybe we can have `reduce`-ed data as constant first, or do some pipe()-like solution.

src/components/table/table.tsx
```diff
@@ -113,16 +128,29 @@ export function Table<T extends object>({
...
+        selectedFlatRows.reduce((h: Item[], r: Row<T>) => {
```
Azein, line 141: Do `h` and `r` represent some naming convention? We generally use `acc` and `current`/`entity name from context`.

src/components/table/table.tsx
```diff
@@ -113,16 +128,29 @@ export function Table<T extends object>({
...
+          if (isGrouped && r.isGrouped) return h
+          if (itemIsDisabled(r)) return h
+          return [...h, r.original]
```
Azein, line 144: Nothing wrong with this, however it's safe to use `push` in potentially lengthy iterations, to not re-create the array. I think it would be faster, yet less idiomatic.

src/components/table/table.tsx
```diff
@@ -113,16 +128,29 @@ export function Table<T extends object>({
...
+        }, [])
+      )
     }
-  }, [selectedFlatRows, selectedItemsClb])
+  }, [selectedFlatRows, isAllRowsSelected, selectedItemsClb, groupBy, itemIsDisabled])
+
+  useEffect(() => {
```
Azein, line 150: I'm a bit confused with this effect. If we get `isAllRowsExpanded`, and it is falsy, we make them `expanded` by an imperative command, right? Why?

src/components/table/components/table-row/table-row.tsx
```diff
@@ -23,7 +23,7 @@ const RowLayout = ({ children, layoutType, ...props }: any) => {
 const DefaultGroupHead = ({ row, layoutType, style }: any) => {
   const rowProps = row.getRowProps()
   return layoutType === "table" ? (
-    <tr {...rowProps} style={style}>
+    <tr title="group-head" {...rowProps} style={style}>
```
Azein, line 26: Is it some a11y/semantic requirement?

novykh: actually it's just for testing purposes, but yes accessibility wise it's better than not having anything :)

Azein:
Code is good to go, however this is a breaking change, and preparations should be made across app tables that use selection.

src/components/table/virtualized-table.tsx
```diff
@@ -125,15 +132,30 @@ export function VirtualizedTable<T extends object>({
...
+      }, [])
+
+      selectedItemsClb(validRows)
     }
-  }, [selectedFlatRows, selectedItemsClb])
+  }, [selectedFlatRows, isAllRowsSelected, selectedItemsClb, groupBy, itemIsDisabled])
+
+  useEffect(() => {
+    if (isAllRowsExpanded) return
+    toggleAllRowsExpanded()
```
jacekkolasa, line 157: can you please use brackets here? or prettify formatted it like this?

jacekkolasa (approved):
lgtm! approving, but i left a small styling suggestion

jacekkolasa (approved):

PR #96, opened 2020-07-03
Introduced tabs components.

novykh:
Created tabs and tab components.

Tabs can work as controlled or uncontrolled.
Tab can be disabled.

Added stories to play around.
Added tests for various states and actions.

Future:
We can change the border bottom of an active tab, from css style to an actual element that animates across the tabs on change.
We can change the panel render logic when tabs are being used with a router.


Azein:
I have a suggestion to change the approach - use key schema, instead of array/indexes. After working with changing feature flags for components using Tab (our current one also is array-based), I think it would be better. Indexes are dynamic and not reliable because people can turn on and off the flags in localStorage while testing the feature - so you have either to produce overloaded conditions, or placeholder invisible tabs to maintain indexes.

Being able to do render conditions like `activeTab === "nodes" &&` helps a lot.

src/components/tabs/tabs.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+const TabContentExample = props => <i>all in italics: {props.children}</i>
+
+const makeTabContentSelect = {
+  TabContentExample,
+  b: "b",
+}
+
+tabsStory.add(
+  "Tabs",
```
Azein, line 33: Let's discuss a case, where we have a limited space for a component using tabs - like, 720px. And the quantity/width of tab "headings" takes more width.  Will tabs "scale", with ellipsis-cutting text, or offer a horizontal scroll?

src/components/tabs/tabs.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+import React, { useState } from "react"
+import "@testing-library/jest-dom/extend-expect"
+import "jest-styled-components"
+import { fireEvent } from "@testing-library/react"
+import { Tabs, Tab, TabsProps } from "."
+import { DefaultTheme } from "../../theme/default"
+import { testWrapper } from "../../../test-utils"
+
+describe("Tabs states", () => {
```
Azein, line 13: Thanks for the tests effort =)

src/components/tabs/tab.tsx
```diff
@@ -0,0 +1,27 @@
...
+  disabled?: boolean
+}
+
+interface TabWrapperProps extends TabProps {
+  label: string | JSX.Element
+  onChange?: OnChange
+  index?: number
+}
+
+export const Tab: FC<TabWrapperProps> = ({ label, index, onChange, ...rest }: TabWrapperProps) => {
```
Azein, line 19: FC<> usage actually isn't a requirement ) We should remove it from readme.

src/components/tabs/tabs.tsx
```diff
@@ -0,0 +1,38 @@
...
+  children?: ReactNode
+  TabContent?: string | React.ComponentType<any>
+}
+
+export const Tabs: FC<TabsProps> = ({
+  className,
+  onChange,
+  selected,
+  children,
+  TabContent = "div",
```
Azein, line 18: Is this prop actually needed? Content itself could be some react node, right?

novykh: Actually it's kind of an important one 😄 
Each active tab is getting rendered inside this tag, this means we can use a layout component without having to duplicate the code for each one.

Example:
```
const TabContent = props => <div style={{padding: 50, backgroundColor: red}}>{props.children}</div>

...
<Tabs TabContent={TabContent}>
  <Tab label="hi">Hello</Tab>
  <Tab label="Hi again">Hello again</Tab>
  <Tab label="Bye">Goodbye</Tab>
  <Tab label="Goodbye">Fairwell</Tab>
</Tabs>
```

Azein:
```
<Tabs>
  <Tab label="hi">
    <StyledTabComponent />
</Tab>
  <Tab label="Hi again">
     <AnotherComponent />
  </Tab>
</Tabs>
```
Usually tabs are more than just text, and they will have own container anyway, so we can simply remove this layer of control from Tabs component. I feel like this additional responsibility can be easily taken to the userland.

novykh: I can use `React.Fragment` as default. Let's see if that's ok

Azein: Let's do it like that then.

novykh: Already pushed it ;)
Let's merge it!

PR #101, opened 2020-08-13
React filter box

Azein:
Filter component based on `react-filter-box`. Extends some functionality for our needs.
Also adds a callback prop to the table components to access/count filtered results.
 Test is not functional, as I need to dig the underlying libraries problems with running in jsdom. It might take a long time, so postponing it atm.


src/components/filter-box/filter-box.tst.tsx
```diff
@@ -0,0 +1,58 @@
...
+        <div key={name}>
+          <Text className="contentClass">{`${name} ${position}`}</Text>
+        </div>
+      ))}
+    </>
+  )
+}
+
+describe("Filter Box test", () => {
+  it(" * should filter results by expression", () => {
```
jacekkolasa, line 50: @Azein is the `tst` in filename a typo, or a way to skip tests? :) we can also do `describe.skip("...` or `it.skip("...)`

Azein: I mentioned it in PR description. Test doesn't work as something inside react-filter-box, maybe the code editor dependency, is breaking in jsdom and it will take time to figure out the solution.

PR #106, opened 2020-08-25
Allow button to work with other tags.

novykh:
- As per styled-components specification, using `as` prop, we should allow the override of the default tag.
- Main usage for this, would be with the router Link component.


VLegakis (approved):
LGTM

PR #108, opened 2020-08-26
collapse icon added

builat:


src/components/icon/assets/collapse.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="16" height="2" viewBox="0 0 16 2" xmlns="http://www.w3.org/2000/svg">
```
VLegakis, line 1: Seems ok. I just have an impression we have a static/assets folder for assets. Not sure 100% but seems strange having them here in the src.

VLegakis (approved):
LGTM. Just verify this static/assets folder since I don't remember very well. E.g how are the assets packaged and served etc..

jacekkolasa (approved):

PR #110, opened 2020-08-28
Dark mode and semantic-ish colors

novykh:


VLegakis (approved):

src/components/button/button.tsx
```diff
@@ -3,7 +3,8 @@ import { StyledButton } from "./styled"
 import { Icon } from "../icon"
 import { LoaderIcon } from "../icon/components"
 
-export type ButtonType = "default" | "hollow" | "borderless"
+export type ButtonFlavour = "default" | "hollow" | "borderless"
+export type ThemeType = "light" | "dark"
```
jacekkolasa, line 7: should Theme be imported from other, common place?

src/components/button/styled.tsx
```diff
@@ -1,47 +1,66 @@
...
-      borderActive: getSpecialColor || getGreenMalachite,
+      color: getTextColor,
+      colorHover: getTextColor,
+      colorActive: getTextColor,
+      bg: getSpecialColor || getPrimaryColor,
+      bgHover: getSpecialColor || getPrimaryColor,
+      bgActive: getSpecialColor || getAccentColor,
+      border: getSpecialColor || getPrimaryColor,
+      borderHover: getSpecialColor || getAccentColor,
+      borderActive: getSpecialColor || getAccentColor,
```
jacekkolasa, line 47: +1

src/components/button/button.tsx
```diff
@@ -18,10 +19,13 @@ export interface ButtonProps {
   onMouseUp?: (e: SyntheticEvent<HTMLButtonElement>) => void
   danger?: boolean
   warning?: boolean
-  type?: ButtonType | undefined
-  flavour?: ButtonType
+  type?: string
```
jacekkolasa, line 22: how about this?
```suggestion
  type?: React.ButtonHTMLAttributes<Element>["type"]
```

jacekkolasa (commented):
so glad we'll have those semantic colors! good stuff! few comments

src/components/button/button.tsx
novykh: Of course! This is just for the cases where we have a button on a dark background, and we need to override the button theme in place. In general, the only way to change the theme is through the main themeProvider mechanism.

The main example for overriding needs is when we have a simple borderless button on a black background, where we need it to work always like it's in "dark" mode. Otherwise, grey text won't look good on black background :)

jacekkolasa: that's good to know, thx!

PR #112, opened 2020-08-30
fill attr removed from pin icon

builat:


jacekkolasa (approved):
LGTM

novykh (commented):
I'd suggest to not add static fill/color attrs. Can't you pass the color you need in place when you render it?

builat:
@novykh this icon exported from figma. For now we have to remove this attributes by hand. Some time ago we discussed svgr based util for this.

PR #116, opened 2020-09-02
dynamic data update added

Azein:


builat (approved):

src/components/filter-box/extended-filter-box.tsx
```diff
@@ -0,0 +1,12 @@
+import ReactFilterBox from "react-filter-box"
+
+// Library doesn't handle dynamic data, even given an updated handler instance in props
+// So we extend base component and update its parser manually
+// to enable autosuggestions for dynamic data
+export class ExtendedFilterBox extends ReactFilterBox {
+  componentDidUpdate(prevProps: any) {
+    if (prevProps.autoCompleteHandler !== this.props.autoCompleteHandler) {
```
VLegakis, line 8: Not sure about what this is doing, but shouldn't this call its parent. 
The way it is seems we completely override it

VLegakis (approved):
Not sure what this is doing but I suppose it should call its parent.
If this has been taken care of please ignore the comment

PR #124, opened 2020-09-09
Tabs support for feature-flagged parts

Azein:
In expressions like
```
{isFeatureAvailable() && <Tab ..... />}
```
value `false` will be included in React children array, which messes up useBuildTabs hook, resulting in unreachable feature-flagged tabs, if there are multiple of them and some are enabled and some not.

This PR offers a solution to allow usage of such simple expressions as mentioned above, with persisting indices without actually including "false" part into navs.

### DIFF
--- a/src/components/tabs/tabs-hooks.tsx
+++ b/src/components/tabs/tabs-hooks.tsx
@@ -16,14 +16,15 @@ export const useBuildTabs = (
     let activeIsDisabled: boolean = false
 
     Children.forEach(children, (tab, index) => {
-      if (!tab) return
-      const { props } = tab
+      const props = tab?.props || {}
       if (firstActiveIndex < 0 && !props.disabled) firstActiveIndex = index
 
       const isActive = activeIndex === indeces.length
       const key = `${index}-${props.label}`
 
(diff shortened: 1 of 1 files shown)


builat (approved):

src/components/tabs/tabs-hooks.tsx
```diff
@@ -16,14 +16,15 @@ export const useBuildTabs = (
     let activeIsDisabled: boolean = false
 
     Children.forEach(children, (tab, index) => {
-      if (!tab) return
-      const { props } = tab
```
novykh, line 20: I dont see any difference with the changes you made. Line 19 that you removed returns fast for falsy tabs and do not push children. Can you elaborate more about this change?

novykh (changes requested):
By looking at it (currently on my phone, will check it thoroughly tomorrow) both implementations seem to do the same thing. Not pushing falsey tabs to the children array. Thus, I dont think there is any need to change this, unless of course there is indeed a problem. Can you provide a test case maybe failing with the first implementation but passing with the second? Lets talk tomorrow

novykh (approved):

PR #131, opened 2020-09-23
Filter box layout

Azein:
Added inline mode for effortless vertical centering, placeholder, meta info display modes and ability to pass ReactNode for results quantity.

### DIFF
--- a/src/components/filter-box/README.md
+++ b/src/components/filter-box/README.md
@@ -71,5 +71,7 @@ interface Props {
   onFocus?: () => void
   onBlur?: () => void
   editorConfig?: Object
+  inline?: boolean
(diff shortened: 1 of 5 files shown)


builat (approved):

VLegakis (approved):

VLegakis:
Can you explain what do you mean you can pass ReactNode for results quantity param?

Azein:
@VLegakis in figma, they have custom text to filter results quantity, like `10 Nodes`, not just `10` as in previous design iteration, and it also has text color for part of the text etc, so I decided that ability to pass just a ReactNode is the simple way to go here. This way colors/styles and plurals approach can be decided on the user side.

PR #142, opened 2020-11-10
modal overlays position:fixed

jacekkolasa:
PortalSidebar and Overlay have position: absolute, but we need position: fixed on Dashboard, where <html /> is scrollable.

I've checked on cloud if it doesn't break anything but it looks good

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/netdata-ui",
-  "version": "0.10.33",
+  "version": "0.10.34",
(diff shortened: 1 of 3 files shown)


VLegakis (approved):
Nice! I’ll try to test it a bit also for the rest of uses in the cloud. Many thanks Jacek!

novykh (approved):

apostolidhs (approved):

VLegakis (approved):

allelos (approved):

PR #143, opened 2020-11-11
add arrow left icon

allelos:


src/components/icon/assets/arrow_left.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
novykh, line 1: I think you should remove both (on `<svg>` and `<path>`) the static `fill` props

novykh (approved):

PR #145, opened 2020-11-14
Updated filter box to work with initial value.

novykh:
Passing initial value to filter box should be passed to code-mirror editor and also, parse the query to return any valid expressions.
- Fixed prop delegation from our component to library.
- Extend library's component to parse query (if any) on mount.

### DIFF
--- a/src/components/filter-box/extended-filter-box.tsx
+++ b/src/components/filter-box/extended-filter-box.tsx
@@ -4,6 +4,16 @@ import ReactFilterBox from "@netdata/react-filter-box"
 // So we extend base component and update its parser manually
 // to enable autosuggestions for dynamic data
 export class ExtendedFilterBox extends ReactFilterBox {
+  componentDidMount() {
+    if (super.componentDidMount) {
+      super.componentDidMount()
+    }
+    if (this.props.query) {
+      const result = this.parser.parse(this.props.query)
+      if (this.props.onParseOk) this.props.onParseOk(result)
+    }
+  }
+
   componentDidUpdate(prevProps: any, prevState: any) {
     if (super.componentDidUpdate) {
       super.componentDidUpdate(prevProps, prevState)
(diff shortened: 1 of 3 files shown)


src/components/filter-box/extended-filter-box.tsx
```diff
@@ -4,6 +4,16 @@ import ReactFilterBox from "@netdata/react-filter-box"
 // So we extend base component and update its parser manually
 // to enable autosuggestions for dynamic data
 export class ExtendedFilterBox extends ReactFilterBox {
+  componentDidMount() {
+    if (super.componentDidMount) {
```
apostolidhs, line 8: Which is the case that we don't have a super instance?

apostolidhs (approved):

src/components/filter-box/extended-filter-box.tsx
VLegakis: Correct @apostolidhs ! I think author was defensive here. Lets remove this check

src/components/filter-box/extended-filter-box.tsx
```diff
@@ -4,6 +4,16 @@ import ReactFilterBox from "@netdata/react-filter-box"
...
+      super.componentDidMount()
+    }
+    if (this.props.query) {
+      const result = this.parser.parse(this.props.query)
+      if (this.props.onParseOk) this.props.onParseOk(result)
```
VLegakis, line 13: We should better use bracketed conditionals if possible. 
Both for maintainability and readability...
https://stackoverflow.com/questions/2125066/is-it-a-bad-practice-to-use-an-if-statement-without-curly-braces

VLegakis (approved):
Nice! Only some optional comments

src/components/filter-box/extended-filter-box.tsx
novykh: Since this is prototypal inheritance, we need to check that filter-box component has a `componentDidMount`. Otherwise there is a chance on a patch upgrade of the library that removes its own `componentDidMount` to break our code.

src/components/filter-box/extended-filter-box.tsx
novykh: The problem with readability lies on the `if..else`. One liner if is fine afaic.

PR #148, opened 2020-11-19
Yarn and upgrades

novykh:
- Upgraded all the packages to latest.
- Using yarn, since it's faster with better cache than npm itself.
- Also, with yarn we will transition better in the future when we will introduce workspaces (to include all of our open source projects, under one monorepo)
- Last, added path resolutions for `src` and `utils`.


apostolidhs (approved):
YESSSS 🚀

allelos (approved):

package.json
```diff
@@ -8,37 +8,35 @@
...
-    "@rmwc/button": "^5.7",
-    "@rmwc/circular-progress": "^5.7",
-    "polished": "^3.4.1",
-    "ramda": "^0.26.1",
+    "polished": "^4.0.3",
+    "ramda": "^0.27.1",
     "react-portal": "^4.2.0",
     "react-table": "^7.2.1",
-    "react-use": "^12.7.1",
+    "react-use": "^15.3.4",
```
jacekkolasa, line 33: I think there are some breaking changes in this lib (maybe in others, too), but perhaps netdata-ui will just use different version until we update it also in Dashboard, right?

jacekkolasa (approved):
LGTM, but have we checked if it doesn't break Dashboard? I'm worried what will happen if it's breaking something and we'll need to add new icons or any other small changes

package.json
novykh: yep. I think there are only new hooks in it tho.

novykh:
@jacekkolasa how do you feel about bumping a major version on this one? But this means if dashboard needs ui kit changes, should **start a branch from the tag it uses**.
@netdata/cloud-fe thoughts?

package.json
jacekkolasa: you're right, i went through the list and any breaking changes were related to hooks we don't use.

jacekkolasa:
>  But this means if dashboard needs ui kit changes, should start a branch from the tag it uses.
@novykh sure, good idea, just to have this options as a backup plan in case something broke

VLegakis (approved):

PR #150, opened 2020-11-20
Flex component

apostolidhs:
Will be merged once https://github.com/netdata/netdata-ui/pull/148 merged on master

A general purposes Flexbox container.
Builds most of the layout needs without the need of extending the styled components.

Screenshot 2020-11-20 at 4 38 26 PM
Screenshot 2020-11-20 at 4 39 22 PM
Screenshot 2020-11-20 at 4 38 54 PM

### DIFF
--- a/src/components/templates/flex/index.stories.js
+++ b/src/components/templates/flex/index.stories.js
@@ -1,0 +1,83 @@
+import React, { useState } from "react"
+import { storiesOf } from "@storybook/react"
+import { readmeCleanup } from "utils/readme"
+import { H4, Text, TextSmall } from "src/components/typography"
+// @ts-ignore
+import readme from "./README.md"
+
+import Flex from "./index"
+
+const flexStory = storiesOf("COMPONENTS|Templates/Flex", module)
+
+const subData = {
+  readme: {
+    sidebar: readmeCleanup(readme),
+  },
+  jest: ["index.test.js"],
+}
+
+const Card = () => (
+  <Flex
+    height="260px"
+    width={{ max: "220px" }}
+    background="disabled"
+    padding={[4]}
+    margin={[0, 0, 4, 0]}
+    gap={4}
+    round
+    column
+  >
+    <Flex basis="50%" background="placeholder" />
+    <H4 as="span">This is a card title</H4>
+    <Text>This is a card description that contains useful information</Text>
+    <TextSmall>April, 1997</TextSmall>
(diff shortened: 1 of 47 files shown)


src/components/templates/flex/index.stories.js
```diff
@@ -0,0 +1,83 @@
...
+  },
+  jest: ["index.test.js"],
+}
+
+const Card = () => (
+  <Flex
+    height="260px"
+    width={{ max: "220px" }}
+    background="disabled"
+    padding={[4]}
```
VLegakis, line 24: I suppose this is pixels so this defeats the 8px (or 4px) breakpoint system we discussed. But I guess that would be ok for now.

src/components/templates/flex/index.stories.js
```diff
@@ -0,0 +1,83 @@
...
+  ),
+  subData
+)
+
+flexStory.add(
+  "Header",
+  () => (
+    <Flex
+      as="header"
+      justifyContent="between"
```
VLegakis, line 72: So this can also be `around` etc? Are these somewhere documented (or even better could we better use the original CSS `space-between` etc?)

src/components/templates/mixins/alignContent.js
```diff
@@ -0,0 +1,11 @@
+const alignContentMap = {
```
VLegakis, line 1: Ah here they are! Ok :)

VLegakis (approved):
Great work @apostolidhs !
Something I may have missed: would it make sense to also have a `disabled` prop for the whole Flex div? (opacity 0.3 and pointer-events:none)?
Not sure about that just an idea which may prove useful (maybe you already predicted that and I missed it)

src/components/templates/flex/index.stories.js
apostolidhs: padding, margin and gap are multipliers of `SIZE_SUB_UNIT` = 4px. 
So, the `padding={[4]}` evaluates to `padding: 16px;`

apostolidhs:
> Great work @apostolidhs !
> Something I may have missed: would it make sense to also have a `disabled` prop for the whole Flex div? (opacity 0.3 and pointer-events:none)?
> Not sure about that just an idea which may prove useful (maybe you already predicted that and I missed it)

The is not any related pattern in the cloud repository, so I don't think that the `disabled` property you described is reusable.

PR #151, opened 2020-11-23
Refactored typography

apostolidhs:
The upgrade does not break the previous clients because all the new properties are optional or default to the previous values (see `font-color`).

The typography components will have the ability to:

* fontColor
* alignSelf
* textAlign
* textTransform
* truncate
* wordBreak
* margin
* padding

The typography elements are also composable exposing the factory functions. Example:

```
const Heading = makeH1('div'); // I expect that this will never happen
```


VLegakis (approved):
Nice! Only a note do we need to export these `makeH4` etc functions? Aren't they just for internal UIKit usage?

apostolidhs:
> Nice! Only a note do we need to export these `makeH4` etc functions? Aren't they just for internal UIKit usage?

Those factories are useful for setting the typography flavor on a component. As you say this will not happen frequently, but we need it. For instance,

```
import {OptionItemHeader} from 'react-select'
const StyledOptionItemHeader = makeH4(makeFlex(OptionItemHeader))
```

PR #155, opened 2020-11-23
Resolved aliases after typescript build.

apostolidhs:
**Before**
The emitted files in the `lib` where resolved with the alias path, so the clients could not import them.

**After** 
The emitted files contain the relative paths.


novykh (approved):

novykh:
also `yarn version --patch`

PR #157, opened 2020-11-25
Icons

VLegakis:


src/components/icon/assets/node_child.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="16" height="13" viewBox="0 0 16 13" xmlns="http://www.w3.org/2000/svg">
```
allelos, line 1: All of these icons have different `viewBox` but in my opinion they should be consistent. For example `24 x 24` in order for this to work correctly and not have inconsistencies we should probably be exporting the container with the icon from figma and not only the vector.

VLegakis: Thanks for that @allelos !
Not sure about that since we provide the container. but in any case I pushed them with the container (which also seems to have different sizes per icon) and let's see if it suits us better.

jacekkolasa (approved):

allelos (approved):

PR #160, opened 2020-11-30
update icon list

allelos:


VLegakis (approved):

src/components/icon/icons-list.ts
```diff
@@ -68,6 +69,7 @@ import logo_s from "./assets/logo_s.svg"
 import pencil from "./assets/pencil.svg"
 import chevron_left from "./assets/chevron_left.svg"
 import view_list from "./assets/view_list.svg"
+import view_list_alt from "./assets/view_list_alt.svg"
```
jacekkolasa, line 72: looks like that icon is missing in iconsList?

jacekkolasa (approved):

src/components/icon/icons-list.ts
allelos: actually alt version are removed altogether

PR #179, opened 2021-03-18
Added search in documentation.

novykh:


src/organisms/documentation/search/provider.js
```diff
@@ -0,0 +1,54 @@
...
+    },
+  },
+  alwaysSearchOnInitialLoad: false,
+}
+
+const SearchProvider = ({ children }) => {
+  return (
+    <ElasticSearchProvider config={config}>
+      <WithSearch
+        mapContextToProps={({ searchTerm, setSearchTerm, results, reset }) => ({
```
apostolidhs, line 41: Could be defined in a constant function.

src/organisms/documentation/search/results.js
```diff
@@ -0,0 +1,93 @@
...
+
+const SearchResults = ({ results }) => {
+  const tabbedResults = useTabbedResults(results)
+
+  return (
+    <Flex overflow={{ vertical: "auto" }} data-testid="searchResults" flex>
+      <StyledTabs>
+        {keys.map(key => {
+          const tabResults = tabbedResults[tabValuesByKey[key]]
+          if (!tabResults?.length) return
```
apostolidhs, line 72: This filtering could be moved out in a useMemo section, so we do not return empty elements during the rendering.

src/organisms/documentation/search/results.js
```diff
@@ -0,0 +1,93 @@
...
+const Link = styled.a`
+  text-decoration: none;
+`
+
+const Section = ({ title, description, url }) => (
+  <Flex gap={2} column>
+    <Text strong>{title?.snippet || "Untitled"}</Text>
+    {!!description && <Text>{description?.snippet || "No description"}&hellip;</Text>}
+    <Flex alignSelf="end">
+      <Text as={Link} color="primary" href={url} target="_blank">
```
apostolidhs, line 16: If the `url={url?.raw}` can be falsy, we could handle it without providing an invalid link

apostolidhs (approved):

src/organisms/documentation/search/results.js
novykh: You mean return the actual keys back that have values?

src/organisms/documentation/search/results.js
novykh: it can't be I will remove the `?`

PR #184, opened 2021-04-14
Add support for discourse search

allelos:


novykh (approved):

src/organisms/documentation/search/discourse/index.js
```diff
@@ -0,0 +1,18 @@
...
+
+const DiscourseSearch = ({ children, ...rest }) => {
+  const [topics, setTopics] = useState([])
+
+  const { searchTerm, results } = rest
+
+  useEffect(() => {
+    if (!searchTerm) return
+    const onSuccess = ({ data }) => setTopics(data)
+    fetchTopics(searchTerm, onSuccess)
```
jacekkolasa, line 12: @allelos is it possible that we'll have race-condition problems here? If there will be 2 calls and the first one will resolve after the second one?

allelos: Then the first one will be the final result. 😛 
Well the search input is debounced, not that it totally prevents the race condition you are referring to, but in my opinion this edge case will not be a big problem with this specific feature / addition. If it becomes an occuring problem we can tackle it.

jacekkolasa: If it's debounced, then indeed this is not a big problem..

jacekkolasa (approved):

PR #189, opened 2021-05-05
Reusable pill component

allelos:
**Solid Pills**
image

**Hollow Pills**
image

**Icon Pills**
image

**Custom Pill**
image

**Tiny Pill**
image


src/components/pill/container.js
```diff
@@ -0,0 +1,28 @@
...
+    side: "all",
+    color: borderColor ? borderColor : ["pill", hollow ? "border" : "background", flavour],
+    size: "1px",
+  },
+  height: "20px",
+  justifyContent: 'center',
+  alignItems: 'center'
+}))`
+  ${getBackground};
+  cursor: pointer;
```
novykh, line 25: Can we have a prop for that? `clickable` or something

src/components/pill/index.js
```diff
@@ -0,0 +1,34 @@
+import React, { forwardRef, useCallback } from "react"
+import { TextMicro } from "src/components/typography"
+import { Icon } from "src/components/icon"
+import Container from "./container"
+
+const Pill = forwardRef(({ children, background, color, hollow, icon, reverse, ...rest }, ref) => {
+  const { flavour } = rest
+
+  const renderIcon = useCallback(() => {
```
novykh, line 9: Isn't it better to put it out of the Pill as its own component?

novykh (approved):
I'm not sure that I agree with the pill colors being set on every theme since they are the same across all of them. I would put them inside the component file or on another color file that get's merged in every theme like the rawColors currently do.

src/components/pill/index.js
allelos: Yeap, I totally agree 💯

allelos:
> I'm not sure that I agree with the pill colors being set on every theme since they are the same across all of them. I would put them inside the component file or on another color file that get's merged in every theme like the rawColors currently do.

I don't disagree, a color mapping inside the component was my first approach too. But then I thought that the colors are the same now, tomorrow might be a different case.

jacekkolasa:
Will we need also borderless?
image

src/components/pill/container.js
```diff
@@ -0,0 +1,28 @@
...
+}
+
+const Container = styled(Flex).attrs(({ round = 999, hollow, flavour, borderColor }) => ({
+  padding: [1, 2],
+  round,
+  border: {
+    side: "all",
+    color: borderColor ? borderColor : ["pill", hollow ? "border" : "background", flavour],
+    size: "1px",
+  },
```
jacekkolasa, line 19: Will we need also borderless?
image

src/components/pill/container.js
```diff
@@ -0,0 +1,28 @@
...
+  height: "20px",
+  justifyContent: 'center',
+  alignItems: 'center'
```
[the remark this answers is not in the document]
allelos: Hmmm, I thought [these](https://www.figma.com/file/28Xv1J95zLbPowJujLuArZ/%F0%9F%93%A6Netdata-Design-System?node-id=6243%3A0) were all the pills.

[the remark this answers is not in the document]
allelos: As far as the circular ones, If we remove the `height` and make the `radius` `100%` it will be circular for one digit / char. Let's figure all these out.

src/components/pill/container.js
allelos: I would keep the border, to keep all the pills consistent, but just make it to the `hollow` color when `borderless` prop is present. wdyt?

jacekkolasa (approved):

PR #190, opened 2021-05-06
Updated drops to be able to adapt their positioning so they do not hide their target.

novykh:
Allowed a drop container to not hide the target element if needed.


apostolidhs (approved):
🖼️

allelos (approved):

jacekkolasa (approved):

PR #195, opened 2021-06-07
cursor mixin

jacekkolasa:


allelos (approved):

src/mixins/cursor.js
```diff
@@ -0,0 +1,16 @@
...
+  none: "none",
+  text: "text",
+  wait: "wait",
+  initial: "initial",
+  inherit: "inherit",
+}
+
+export default ({ cursor }) => {
+  if (!cursor) return ""
+  return cursor in cursorMap ? `cursor: ${cursor};` : ""
```
allelos, line 15: We could also check if there is `onClick` prop present

jacekkolasa: @allelos interesting idea, but isn't it a little "magical"? I have no opinion, how about we discuss it on team meeting? This is generally a question of how low-level our components should be.

I'm a little afraid to automatically adjust cursor for all Flex components without testing - it could break for example when component is disabled but onClick handler is still there.

allelos: We can have `cursor` prop override other cases when present. But in any case, let's keep it for a future iteration :)

PR #199, opened 2021-06-15
getTransparentBackground for hollow button

jjtsou:


jacekkolasa (approved):
approving on assumption that it was tested in all use cases :)
And i think tests need to be updated

PR #215, opened 2021-07-20
add text-decoration to typography components

jacekkolasa:


allelos (approved):

allelos (commented):
Could we also add to the [index.d.ts](https://github.com/netdata/netdata-ui/blob/master/src/components/typography/mixins/index.d.ts) and [typography index.d.ts](https://github.com/netdata/netdata-ui/blob/master/src/components/typography/index.d.ts) file so we can have props available on autocomplete etc?

jacekkolasa:
@allelos absolutely! i forgot about those files

PR #217, opened 2021-08-12
Feature/onboarding/claiming nodes

jjtsou:


src/components/tabs/tab.js
```diff
@@ -1,12 +1,12 @@
...
 import { StyledTab } from "./styled"
 
-export const Tab = ({ label, index, onChange, ...rest }) => {
+export const Tab = ({ index, onChange, ...rest }) => {
   const onClick = useCallback(() => onChange && onChange(index || 0), [index, onChange])
 
   return (
     <StyledTab onClick={rest.disabled ? undefined : onClick} {...rest}>
-      {label}
+      {rest.label}
```
jacekkolasa, line 9: out of curiosity, this change will pass `label` prop to StyledTab, is this intentional?

jacekkolasa (approved):

jacekkolasa:
@jjtsou i published it so it can be used in https://github.com/netdata/cloud-frontend/pull/2682

src/components/tabs/tab.js
jjtsou: Yes, this change was intentional. [For reference](https://github.com/netdata/cloud-frontend/pull/2681/files#diff-052cc627f52854365d9eba7cdbb3ab0fd1a9bb845e3d1b8c328c37467c0cfac9)

jjtsou: ` <Tab label={<TabLabel label="linux" />} as={StyledButton}>`
Button takes it's label as a prop.

jacekkolasa: @jjtsou thanks!

PR #222, opened 2021-09-15
Onboarding pt1 icons

jacekkolasa:
- exclamation icon (for onboarding)
- two arrow icons (for filter bar)


src/components/icon/assets/arrow_w_line_right.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="24" height="13" viewBox="0 0 24 13" xmlns="http://www.w3.org/2000/svg">
```
allelos, line 1: Can we get the svg with the container from Figma so it has a consistent box? `24 x 24` for example

novykh (approved):

PR #244, opened 2021-12-23
Update the forms and textfields

AlexNti:
This PR introduces the new UI/UX designs for checkboxes and radio buttons.
https://app.zenhub.com/workspaces/product-5ea1534e9ff5df7daace00d5/issues/netdata/product/2296

### DIFF
--- a/src/components/checkbox/styled.js
+++ b/src/components/checkbox/styled.js
@@ -43,7 +43,7 @@ export const StyledCheckbox = styled.div`
   border-radius: 4px;
   background: ${({ disabled }) =>
     disabled ? getColor("mainBackgroundDisabled") : getColor("mainBackground")};
-  border: 1px solid ${getValidatedControlColor("inputBorder")};
+  border: 1px solid ${getValidatedControlColor("inputBorder", "inputBorder")};
   transition: all 150ms;
 
   ${HiddenCheckboxInput}:focus + & {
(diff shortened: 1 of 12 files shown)


src/components/checkbox/styled.js
```diff
@@ -40,9 +40,10 @@ export const StyledCheckbox = styled.div`
   align-items: center;
   width: 20px;
   height: 20px;
+  border-radius: 4px;
   background: ${({ disabled }) =>
     disabled ? getColor("mainBackgroundDisabled") : getColor("mainBackground")};
-  border: 1px solid ${getValidatedControlColor()};
+  border: 1px solid ${getValidatedControlColor("inputBorder", "inputBorder")};
   transition: all 150ms;
```
burbuli8ra, line 47: It's a bit irrelevant with this PR, but why isn't `Flex` being used here?

burbuli8ra (approved):

src/components/checkbox/styled.js
AlexNti: I totaly agree with you @burbuli8ra , the change is irrelevant but it was the only way to apply the disabled styles. 
As for the Flex the reason i didn't replaced the styled with the Flex is that it will make this PR more complicated. I was thinking as better idea to start a different branch and refactor all the checkbox component.  (https://github.com/netdata/netdata-ui/tree/refactor-checkbox/src/components/checkbox)

PR #248, opened 2021-12-30
Feature/pseudo flex

AlexNti:
In this PR we introduce support of pseudoselectors for flex component.

Flex can receive styles that will applied when a pseudo class is triggered to an element.
How to use it:

`
<Flex _hover={border:{size: "1px",type: "solid",color:"red"}} />
`

Currently we only support **shadow**,**border** and **background** properties but this can be easily extended depending on the needs.

Pseudoselectors styles can also be applied to an element by applying a **data-**  attribute to the element. 
For example if we want to apply **_focus** styles to an element we can pass **data-focus={""}** to that element or if we want to remove those styles we can pass **data-focus={undefined}**.

A good use case for this is when we have a child element that is getting **focus** and we would like focus styles to be applied to the parent as well (or to a sibiling element) we can do the following:

`const [isChildFocused,setChildFocused] = React.useState();
`

`const flexProps = { "data-focus" : isChildFocused ? "" : undefined };
`

`const flexStyles= {_focus:{border:{size: "1px",type: "solid",color:"red"}}}
`

`<Flex {...flexStyles} {...flexProps}>
`    
`<input onBlur={()=>{setChildFocused(false}} onFocus={()=>{setChildFocused(true}}/>
`
`<Flex>
`


burbuli8ra (approved):
🚀

jacekkolasa (approved):

jjtsou (approved):

PR #259, opened 2022-01-17
Feat/dynamically toggle theme

AlexNti:
If ths PR applied it will allow us to dynamically change the theme color at storybook.


burbuli8ra (approved):

jacekkolasa (approved):
That was needed so much, awesome!

PR #261, opened 2022-01-18
Change styles checkbox

AlexNti:
if this PR  applied it will fix the following bug [https://github.com/netdata/netdata-cloud/issues/233](url).
Also we made a small refactor in order to fix that bug.


.storybook/config.tsx
```diff
@@ -12,6 +12,7 @@ import { ThemeProvider } from "styled-components"
 import Flex from "../src/components/templates/flex"
```
AlexNti, line 12: The change at this file are irrelevant with the this issue. 
This change fixes a bug at storybook where labels does not change colour on dark theme.

.storybook/config.tsx
```diff
@@ -58,10 +59,10 @@ addDecorator(story => {
               checked={isDarkTheme}
               colored={false}
               disabled={false}
-              Label={"asdas"}
+              Label={Text}
```
burbuli8ra, line 62: Why is label with capital?

package.json
```diff
@@ -34,6 +34,7 @@
     "@elastic/search-ui-site-search-connector": "^1.5.1",
     "@netdata/react-filter-box": "^1.0.0",
     "@prismicio/client": "^5.1.0",
+    "@testing-library/user-event": "^13.5.0",
```
burbuli8ra, line 37: 🎉

burbuli8ra (approved):

.storybook/config.tsx
AlexNti: hmmm, i believe this is capital to indicate that this input property expects a component. Check `toggle.js `for more information.

PR #262, opened 2022-01-27
chore: update icons list

burbuli8ra:
This PR includes the following:
* Enhance icons list with 2 new icons
* Update `eslintrc` with TS support
* Add missing TS declaration files and update existing ones


src/components/drops/mixins/index.d.ts
```diff
@@ -26,5 +26,5 @@ export interface AlignProps {
 }
 
 export interface StretchProps {
-  streach: boolean | "width" | "align"
+  streach?: boolean | "width" | "align"
```
novykh, line 29: typo

novykh (approved):

src/components/drops/mixins/index.d.ts
burbuli8ra: Nice catch @novykh !
I will open a separate issue for this one, since I need to check it's usage before fixing the typo!

burbuli8ra: Yes, thought it was passed as `streach`, but checked it as well, so I am pushing an update on that! :)

Issue #264, opened 2022-01-29
Update MasterCard component to support leftPill and rightPill objects


burbuli8ra:
Current implementation of `MasterCard` component supports and array of values for `texts`, `onClicks`, etc.
Since the component supports only 2 pills it would be better to get each pill's properties from an object (right and left).

burbuli8ra:
Please note that the fix will also require an update to `cloud-frontend` where the component is being used:
1. Node's table view
2. Single node view
3. Searchable nodes dropdown in header (pending merge)

PR #267, opened 2022-02-01
chore: update MasterCard component

burbuli8ra:
This PR closes issue #264 

#### Solution
Update `MasterCard` component to support pill objects instead of arrays of strings

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/netdata-ui",
-  "version": "1.7.36",
+  "version": "1.7.37",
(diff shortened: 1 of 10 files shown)


novykh (approved):
Maybe is better to change the name of neutral grey and iron flavors to something semantic? I dont know if they are `default` or `disabled` or something else.

AlexNti (approved):

burbuli8ra:
@novykh they are both disabled, but could be `errorDisabled` and `warningDisabled` instead of `neutral{etc}` 😄

PR #271, opened 2022-02-08
warning_triangle hollow icon, more flexible icon rotate

jacekkolasa:
add warning_triangle_hollow icon
allow negative rotate values for the icon (useful when making transitions)

PR #272, opened 2022-02-09
fix: Box export

burbuli8ra:
This PR also includes:
* Update to `Flex` declaration files
* Add `Box` declaration file in `copy-assets`
* Update icon list with `notificationWarning` icon


jacekkolasa (approved):

src/components/templates/box/index.d.ts
```diff
@@ -33,3 +33,9 @@ export type BoxProps<T extends React.ElementType = any> = CombinedStyledProps<T>
   as?: React.ElementType
   sx?: CombinedStyledProps<T>
 }
+
+declare const Box: React.FC<BoxProps & JSX.IntrinsicElements["div"]>
+
```
AlexNti, line 38: We don't need JSX.IntrinsicElements["div"]. Replace it with React.FC<BoxProps<"div">>. Box props is extending `React.ElementType ` which uses JSX.IntrinsicElements under the hood.

AlexNti (approved):

burbuli8ra:
The fix has already been made in PR #271!

PR #276, opened 2022-02-10
chore: add icon forward ref support

burbuli8ra:
This PR also includes the following:
* Update TS declaration files for `MasterCard` and `Pill` components
* Revert fix for position Y of drops, since it causes issues on `Menu` displayed for users in Cloud's Left SideBar
* Update colors

### DIFF
--- a/src/components/drops/drop/useMakeUpdatePosition.js
+++ b/src/components/drops/drop/useMakeUpdatePosition.js
@@ -49,6 +49,8 @@ const getYPosition = (align, targetRect, dropRect, canHideTarget = true) => {
 
   const minY = Math.max(0, y)
 
+  y = Math.min(window.innerHeight - dropRect.height, minY)
+
   if (!canHideTarget && minY !== y)
     return getYPosition(reverseYPosition(align), targetRect, dropRect)
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/netdata-ui",
-  "version": "1.7.42",
+  "version": "1.7.43",
   "description": "netdata UI kit",
   "main": "./lib/index.js",
   "files": [
--- a/src/components/icon/icon.js
+++ b/src/components/icon/icon.js
@@ -1,4 +1,4 @@
-import React from "react"
+import React, { forwardRef } from "react"
 import { iconsList } from "./icons-list"
 import { StyledIcon } from "./styled"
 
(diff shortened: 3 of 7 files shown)


src/components/drops/drop/useMakeUpdatePosition.js
```diff
@@ -49,6 +49,8 @@ const getYPosition = (align, targetRect, dropRect, canHideTarget = true) => {
 
   const minY = Math.max(0, y)
 
+  y = Math.min(window.innerHeight - dropRect.height, minY)
```
novykh, line 52: Which case was it trying to solve in the first place? Getting out of screen.. where?

src/theme/dark/colors.js
```diff
@@ -12,6 +12,7 @@ const appColors = {
   dropdown: rawColors.neutral.outerSpace,
   elementBackground: rawColors.neutral.outerSpace,
   elementBackgroundHover: rawColors.neutral.tuna,
+  headerItemBackground: rawColors.neutral.limedSpruce,
```
novykh, line 15: The naming is off here, maybe `importantNoticeBackground`? or something more specific to what the component is.

novykh (changes requested):

src/theme/dark/colors.js
novykh, line 15: Use one of the other colors that are blackhaze on the light theme.

novykh (changes requested):

src/components/drops/drop/useMakeUpdatePosition.js
burbuli8ra: It was trying to solve the following:
When the screen doesn't have enough height, the dropdown menus of header items render outside of the screen and not at position `y` of the element they are connected to

What was caused by this fix:
Apart from `Drop`, `Menu` also uses the hook, and the menu rendered for user profile, settings, etc. is now misplaced, outside the screen at its bottom

src/theme/dark/colors.js
burbuli8ra: I have talked with the design and told me to use `limed spruce` in dark theme!
Do you need me to re-sync with them and ask them if an alternative color could be used here?

src/components/drops/drop/useMakeUpdatePosition.js
burbuli8ra: I guess you are referring to the issue I was trying to solve in first place, but it does not!
I re-checked it now! 😄 
Since my fix was trying to solve an issue happening when `console` is open, and for all this time there was no issue reported by any of our users, I guess it is better to revert it and open a ticket to add an extra prop to `Drop` component if we need to skip this recalculation!

novykh: yeah, also we dont need to lose time on fixing the app working on really narrow screens 😄

src/theme/dark/colors.js
burbuli8ra: That is my question thought...
There is no current color matching `rawColors.neutral.blackhaze` in default theme and `rawColors.neutral.limedSpruce`, so since I have already synced with the designer on which color to use for dark theme, if you are against of adding an extra color for this case (which I don't get why to be honest), would you like me to do an other sync with the design and ask if it's ok to use one of the existing ones for dark theme, instead of `rawColors.neutral.limedSpruce`?

novykh: I never asked you to sync with the design team. You got a design which was either on light or on dark mode. You have `blackhaze` somewhere, so you use one of the existing colors that we are using across the app.
For example `elementBackground` which makes total sense. You don't ask anyone else in such case.
Now if we see a color that doesn't make sense with the existing colors, then we talk to the designer and we try to find a way to NOT add another color, just because someone told me on slack. We try to figure out if something of our design system lives up to expectations of the designer.

src/components/drops/drop/useMakeUpdatePosition.js
burbuli8ra: Ok! So, I am leaving it as it is!

src/theme/dark/colors.js
burbuli8ra: Ok, thanks! I'll notify design that I will be using one of the existing pairs and remove the commit!

src/components/drops/drop/useMakeUpdatePosition.js
novykh: By that we mean that we keep the previous impl.
This `y = Math.min(window.innerHeight - dropRect.height, minY)` should remain from what I got from our convey. (sorry about writing convey, I was bored to write conversation, but now I wrote more)

novykh (approved):

PR #281, opened 2022-02-15
support StickyHeader property in table

jacekkolasa:


novykh (approved):
Just one question, the react table doesn't have any prop taking care of that?

jacekkolasa:
@novykh i was searching for a config but it looks like there isn't any. There's also https://github.com/GuillaumeJasmin/react-table-sticky library (extension to react-table) but it's not maintained for 2 years and perhaps it would not be compatible with our use-case (AFAIK we use react-table in non-standard way, have some styles overrides etc.)

novykh:
Cool thnx, merge it ;)

burbuli8ra (approved):

PR #282, opened 2022-02-15
Fix double opening dropdown

jacekkolasa:
related to https://github.com/netdata/cloud-frontend/issues/2877

When dropdown is opened and user clicks on dropdown-button (here: `target`), `onClickOutside` captures mousedown which closes the dropdown, and then `click` is captured separately which opens it again.

I added a detection if mousedown was done in dropdown-button, if yes, `onClickOutside` will not trigger.
This will fix dropdown buttons and btw. maybe it will allow us to remove some rmwc buttons

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/netdata-ui",
-  "version": "1.7.44",
+  "version": "1.7.45",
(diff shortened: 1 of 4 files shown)


novykh (approved):
Good job and yay! about removing rmwc :D

PR #283, opened 2022-02-16
Added better handling for tab dimension calculation.

novykh:


src/organisms/navigation/hooks/useOnTabsResize.js
```diff
@@ -33,10 +33,11 @@ export default (parentRef, ref, target, callback, deps) => {
 
     if (!hasChanged(prevValuesRef.current, { width, tabRight })) return
     if (
-      tabRight < width &&
+      tabRight < width - (width - containerRight) &&
```
burbuli8ra, line 36: Maybe it would be better to store the calculation to a constant and use this instead, since we need it in 2 places and for some reason in the future one of the two places might be missed to be updated!

burbuli8ra (approved):

PR #289, opened 2022-02-25
Feat/modal

AlexNti:
If this PR applied it will create a Modal component to the Netdata-ui.
This Modal structure is the same that is being used right now at our cloud-frontend.

### DIFF
--- a/src/components/modal/index.d.ts
+++ b/src/components/modal/index.d.ts
@@ -1,9 +1,13 @@
 import React from "react"
+import { LayerProps } from "../templates/layer"
+import appColors from "../../theme/default/colors"
+import rawColors from "../../theme/rawColors"
 
 export type ModalContentProps = {
   height?: number | string
   width?: number | string
   testId?: string
+  background?: Exclude<keyof typeof appColors, keyof typeof rawColors>
 }
 
 export type ModalHeaderProps = {
@@ -23,7 +27,7 @@ export type ModalCloseButtonProps = {
   testId?: string
 }
 
-declare const Modal: React.FC
+declare const Modal: React.FC<LayerProps>
 declare const ModalContent: React.FC<ModalContentProps>
 declare const ModalHeader: React.FC<ModalHeaderProps>
 declare const ModalBody: React.FC<ModalBodyProps>
--- a/src/components/modal/index.d.ts
+++ b/src/components/modal/index.d.ts
@@ -1,26 +1,26 @@
 import React from "react"
 
 export type ModalContentProps = {
-  height: number | string
-  width: number | string
-  testId: string
+  height?: number | string
+  width?: number | string
+  testId?: string
 }
 
 export type ModalHeaderProps = {
-  testId: string
+  testId?: string
 }
 
 export type ModalBodyProps = {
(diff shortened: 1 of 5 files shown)


src/components/modal/index.d.ts
```diff
@@ -0,0 +1,35 @@
...
+export type ModalHeaderProps = {
+  testId: string
+}
+
+export type ModalBodyProps = {
+  testId: string
+}
+
+export type ModalFooterProps = {
+  testId: string
```
burbuli8ra, line 18: Just a thought here, couldn't we have `testId` (as optional) as a generic type and just import it and extend types where needed?

src/components/modal/index.js
```diff
@@ -0,0 +1,69 @@
...
+      column
+    >
+      {children}
+    </Flex>
+  )
+}
+
+export const ModalHeader = ({ children, testId }) => {
+  return (
+    <Flex data-testid={testId} padding={[6, 6]}>
```
burbuli8ra, line 24:
```suggestion
    <Flex data-testid={testId} padding={[6]}>
```

src/components/modal/index.js
```diff
@@ -0,0 +1,69 @@
...
+      {children}
+    </Flex>
+  )
+}
+
+export const ModalBody = ({ children, testId }) => {
+  return (
+    <Flex data-testid={testId} padding={[0, 6, 0, 6]} column>
```
burbuli8ra, line 32:
```suggestion
    <Flex data-testid={testId} padding={[0, 6]} column>
```

src/components/modal/index.js
```diff
@@ -0,0 +1,69 @@
...
+      {children}
+    </Flex>
+  )
+}
+
+export const ModalFooter = ({ children, testId }) => (
+  <Flex column padding={[0, 6, 0, 6]} flex={1}>
```
burbuli8ra, line 39:
```suggestion
  <Flex column padding={[0, 6]} flex={1}>
```

src/components/modal/index.js
```diff
@@ -0,0 +1,69 @@
...
+    <Flex
+      data-testid={testId}
+      flex={1}
+      padding={[6, 0, 6, 0]}
```
burbuli8ra, line 43:
```suggestion
      padding={[6, 0]}
```

burbuli8ra (approved):

src/components/modal/index.d.ts
AlexNti: Agree with the optional, but what do you mean as generic type?

novykh (approved):

src/components/modal/index.d.ts
novykh: I believe to make have a common interface/type and enhance the component specific interfaces/types?

jjtsou (approved):

src/components/modal/index.d.ts
AlexNti: We can create a  `type PropsWithTest<P> = P & {testId?:string}` this will be simillar on how react is handling props when using `React.FC`

PR #290, opened 2022-02-25
Fix onclick warning

jacekkolasa:
only 1 small change - https://github.com/netdata/netdata-ui/pull/290/commits/1b0683b6f6e3411465422d62b835cbda2932af04

rest if prettify

### DIFF
--- a/src/components/pill/mastercard.js
+++ b/src/components/pill/mastercard.js
@@ -4,29 +4,25 @@ import { getMasterCardBackground } from "./mixins/background"
 import { MasterCardContainer } from "./styled"
 
 const MasterCard = forwardRef(
-  ({
-    children,
-    height,
-    normal,
-    onClick,
-    pillLeft = {},
-    pillRight = {},
-    round,
-    size,
-    ...rest
-  }, ref) => {
+  (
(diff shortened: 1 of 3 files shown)


burbuli8ra (approved):
🎉

src/components/pill/mastercard.js
```diff
@@ -5,7 +5,18 @@ import { MasterCardContainer } from "./styled"
...
+    {
+      children,
+      height,
+      normal,
+      onClick,
+      pillLeft = {},
+      pillRight = {},
+      round,
+      size,
+      "data-testid": dataTestId,
```
burbuli8ra, line 17: Maybe we could have `testId` here to avoid this! :)

burbuli8ra (approved):

src/components/pill/mastercard.js
jacekkolasa: i thought so at beginning, but on the other hand, for all the other components we add prop with dashes, ie.`<Link data-testid="dropdown-node-link"` so if we change the pattern now, you will need to remember which uses which. And this "data-testid": is written only once, so i thought it will be easier just to follow naming we know
WDYT?

burbuli8ra: I'm fine with it, just mentioned it for more convenience! 😄

Issue #265, opened 2022-01-29
Add tests for Pill and MasterCard components


burbuli8ra:
The `Pill` family components do not have any unit tests implemented yet, so we should add some! 😄

PR #297, opened 2022-03-09
chore: update MasterCard component with alert icon

burbuli8ra:
This PR resolves issues #265 and [cloud-frontend #3012](https://github.com/netdata/cloud-frontend/issues/3012)

#### Solution
* Update `MasterCard` component to match the new designs with the alert icon
* Add tests for `Pill` and `MasterCard` components and mixins

Screenshot 2022-03-09 at 14 27 55


src/components/pill/mastercard.js
```diff
@@ -28,27 +41,55 @@ const MasterCard = forwardRef(
...
-            <MasterCardPill side="left" {...pillProps} {...pillLeft} />
-            <MasterCardPill side="right" {...pillProps} {...pillRight} />
+            <MasterCardPill side="left" isAlert={isAlert} data-testid={`${testId}-left-pill`} {...pillProps} {...pillLeft} />
+            <MasterCardPill side="right" data-testid={`${testId}-right-pill`} {...pillProps} {...pillRight} />
           </>
         )}
       </MasterCardContainer>
     )
+
+    if (!isAlert) return MasterCardComponent
```
novykh, line 64: Why do we have this coupling? `Alert` looks like it is a flavour, hence I would expect `Mastercard` to be plain, and use a separate component `AlertMastercard` to accomodate the specifics for alerts. Can we decouple this please?

src/components/pill/mastercardPill.js
```diff
@@ -28,9 +29,10 @@ const MasterCardPill = forwardRef(({ isClickable, flavour, onClick, side, text,
     <Pill
       background={background}
       borderColor={background}
+      data-testid={dataTestId || "mastercard-pill"}
```
novykh, line 32: Those `'data-testid': dataTestId` are not needed and better to not make strict such props. Just pass the rest down the component.

src/components/pill/styled.js
```diff
@@ -1,12 +1,27 @@
...
+  })
+)``
+
+export const MasterCardContainer = styled(Flex).attrs(
+  ({ background, height, isAlert, onClick, round = 999, size }) => ({
+    background,
+    ...(onClick && { cursor: "pointer" }),
     height: getPillHeight(height, size),
-    cursor: onClick ? "pointer" : "default",
+    margin: isAlert && [0, 0, 0, -1],
```
novykh, line 24: Same, this coupling will make it difficult to maintain. You should enhance the simple styles, specifically for the alert in its own component. Or create flavours.

novykh (changes requested):

src/components/pill/mastercard.js
burbuli8ra: Sure! 😄

src/components/pill/mastercardPill.js
novykh:
```
data-testid="static-text"
{...rest}
```
if rest includes a testid it will override the previous/static one.

novykh (approved):
Did it worth it?

PR #302, opened 2022-03-18
chore: update virtualized table

burbuli8ra:
This PR introduces the following updates:
* Update virtualized table to hide headers if needed
* Update virtualized table to render a full width row with cells as children elements or specified inner row
* Update icons list


src/components/table/components/table-row/table-row.js
```diff
@@ -58,18 +74,18 @@ export const TableRow = ({
...
-          <TableCell
-            key={key}
-            cell={cell}
-            selectedRowIds={selectedRowIds}
-            {...cellProps}
-            customProps={customProps}
-          />
-        )
-      })}
+      {Object.keys(groupedRows).length ? (
```
AlexNti, line 77: You can have this in a separate function , in order to be more clear whats going on here. Like the `renderCells` func

src/components/table/utils.js
```diff
@@ -64,3 +64,29 @@ export const getValidRows = ({ selectedFlatRows, isGrouped, itemIsDisabled }) =>
     acc.push(row.original)
     return acc
   }, [])
+
+export const getGroupedCells = cells => cells.reduce((accumulator, cell) => {
```
AlexNti, line 68: Here as well you can create separate function for each of these two cases. Right now it is not very clear what is getting returned in each case.

AlexNti (approved):
Good job :)! Just added some minor comments.

AlexNti (approved):

libekonst (approved):

PR #319, opened 2022-04-18
add highlight area icon

christophidesp:
Add highlight area icon to assets
To be used in Anomalies tab as indication to action.

Developer insight:
There already is the `selected_area.svg` icon, but there is also a `services/selected_area.svg`, which has the same name but is in a different folder. Regardless, it seems that trying to export both results in one of them being overwritten because they have the same name. The `charts` repo bypasses this by directly importing the svg file and rendering it as a `<svg />` element instead of a `<Icon />` component.
Since the agreed convention is to use the `<Icon />` component wherever possible, this is being added as a replacement of the old `selected_area.svg` and an issue for `charts` will be opened to use this instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/netdata-ui",
-  "version": "1.9.17",
+  "version": "1.10.0",
   "description": "netdata UI kit",
   "main": "./lib/index.js",
   "files": [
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/netdata-ui",
-  "version": "1.9.17",
+  "version": "1.10.0",
```
burbuli8ra, line 3: Do we need to do a minor update instead of patching?

libekonst: I think a minor update fits here because this is an addition to the iconList API (although a small one), and not a bugfix.

burbuli8ra: Agreed on that, but it's a tiny change for doing a `minor` update!
Approved by me, up to you! 😛

burbuli8ra (approved):

PR #331, opened 2022-05-03
colors/pie chart colors

jjtsou:


src/theme/default/colors.js
```diff
@@ -52,6 +52,12 @@ const appColors = {
   nodeBadgeBackground: rawColors.neutral.porcelain,
   nodeBadgeBorder: rawColors.neutral.iron,
   nodeBadgeColor: rawColors.neutral.bluebayoux,
+  // PieCharts
+  strokeColorAqua: rawColors.blue.aquamarine,
+  strokeColorIndigo: rawColors.blue.indigo,
```
novykh, line 57: There is no need to add them here, we only add semantic naming in the different themes and reuse them. Since those are the same regardless of the them its better to get them right away from the raw colors.

PR #334, opened 2022-05-06
table/adding-pagination-handlers

jjtsou:


src/components/table/table.js
```diff
@@ -187,8 +189,14 @@ export function Table({
   ])
 
   const showPagination = withPagination && pageCount > 1
-  const goToPreviousPage = useCallback(() => previousPage(), [previousPage])
-  const goToNextPage = useCallback(() => nextPage(), [nextPage])
+  const goToPreviousPage = useCallback(() => {
+    onGoToPrevious()
+    previousPage()
+  }, [previousPage])
```
burbuli8ra, line 195: I understand why `previousPage` needs to be in the deps array and `onGoToPrevious` is not, but are we sure that `onGoToPrevious` will always be the same prop passed and will not change?
Same goes for the next functions!

burbuli8ra (approved):
I have left a comment more for discussing it! 😄

src/components/table/table.js
jjtsou: Hmm. We cannot be sure about it, so I will add them to the deps array

PR #343, opened 2022-06-02
fixing inputs and tab line colors

vinnygats:
Tab line
Screenshot 2022-06-02 at 5 42 17 PM
It is also imported in the drawer, already checked, not so necessary to be so highlighted in that case.

Screenshot 2022-06-02 at 5 41 56 PM
Screenshot 2022-06-02 at 5 42 06 PM
Checked the inputs also in dark mode, they seem okay.

### DIFF
--- a/src/components/input/use-input-styles.js
+++ b/src/components/input/use-input-styles.js
@@ -1,4 +1,3 @@
-import { normalize } from "path"
 import { useMemo, useCallback } from "react"
 
 const makeColor = ({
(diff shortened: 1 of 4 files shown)


burbuli8ra (approved):

jjtsou (approved):

src/components/input/use-input-styles.js
```diff
@@ -1,3 +1,4 @@
+import { normalize } from "path"
```
jjtsou, line 1: Is this needed ?

PR #353, opened 2022-07-05
Feat/space settings netdata table

AlexNti:


src/components/button/IconButton.js
```diff
@@ -0,0 +1,33 @@
...
+      sx={{ width, height }}
+      onClick={onClick}
+      as={Button}
+      flavour={flavour}
+      disabled={disabled}
+      icon={icon}
+      iconColor={iconColor}
+      iconSize={iconSize}
+      {...props}
+    ></Box>
```
burbuli8ra, line 29: 😠

src/components/tableV2/action.js
```diff
@@ -49,25 +50,42 @@ const Action = ({
...
+          <IconButton
+            iconSize="small"
+            data-testid={`netdata-table-action-${id}${testPrefix}`}
+            disabled={disabled}
+            onClick={onActionClicked}
+            icon={icon}
+            flavour="borderless"
+            sx={{
+              width: "16px",
+              height: "18px !important",
```
burbuli8ra, line 86: Is `!important` really needed here?

src/theme/dark/colors.js
```diff
@@ -56,6 +56,8 @@ const appColors = {
   nodeBadgeBackground: rawColors.neutral.limedSpruce,
   nodeBadgeBorder: rawColors.neutral.bluebayoux,
   nodeBadgeColor: rawColors.neutral.white,
+  //Table
+  tableRowBg: rawColors.neutral.shark,
```
burbuli8ra, line 60: I believe we should not use any new colors here, so would it be possible to sync with the design team and use one of the existing ones?

AlexNti: This has been added by design team, @christophidesp

src/components/tableV2/action.js
christophidesp: @burbuli8ra styles were getting override from main button component afair

PR #355, opened 2022-07-19
Feat/new admin setting v1

AlexNti:
If this PR applied it will introduce the ui components and styles that are required for the generalized admin interface.


src/components/button/IconButton.js
```diff
@@ -0,0 +1,34 @@
...
+      onClick={onClick}
+      as={Button}
+      flavour={flavour}
+      disabled={disabled}
+      icon={icon}
+      iconColor={iconColor}
+      iconSize={iconSize}
+      neutral
+      {...props}
+    ></Box>
```
burbuli8ra, line 30: `/>`

burbuli8ra (approved):
@AlexNti this PR contains changes that have been published to `NPM` and I do not think that there are any breaking changes, so I think the best would be to merge this and open a new PR for any other updates needed! 😄

PR #363, opened 2022-08-03
adding replace action to netdata Table

MichaelGamel:


AlexNti (approved):

src/components/tableV2/netdataTable.stories.js
```diff
@@ -317,6 +323,10 @@ StoryTable.add("Full Table functionallity", () => {
   }
 
   const rowActions = {
+    replace: {
+      handleAction: handleDelete,
```
burbuli8ra, line 327: I think this is not the correct action to be triggered here

PR #369, opened 2022-08-23
add tooltip

AlexNti:
This PR it will fix this issue https://github.com/netdata/netdata-cloud/issues/462

### DIFF
--- a/src/organisms/navigation/tab/index.js
+++ b/src/organisms/navigation/tab/index.js
@@ -3,6 +3,7 @@ import { Icon } from "src/components/icon/icon"
 import Flex from "src/components/templates/flex"
 
 import useStylesTab from "./use-styles-tab"
+import Tooltip from "src/components/drops/tooltip"
 
 const Tab = ({
   active,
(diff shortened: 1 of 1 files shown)


jacekkolasa (approved):

src/organisms/navigation/tab/index.js
```diff
@@ -19,6 +20,7 @@ const Tab = ({
   dragHandleProps,
   tabRef,
   showBorderLeft,
+  tooltip = "",
```
burbuli8ra, line 23: Why is this initialization needed?

burbuli8ra (approved):

src/organisms/navigation/tab/index.js
AlexNti: To avoid undefined errors, on the other hand haven't test  it without it

burbuli8ra: Actually I think that if no `tooltip` content is provided by the user, then it should not be visible, that's why I made the question! 😄

PR #378, opened 2022-09-19
remove redundant error and warning declarations

christophidesp:
Remove redundant error and warning declarations
Update dropdown color for dark theme to differentiate it from borders

### DIFF
--- a/src/components/button/button.test.js
+++ b/src/components/button/button.test.js
@@ -188,10 +188,10 @@ describe("Default Button", () => {
   it(" * should render for danger", () => {
     const { container } = renderWithProviders(<Button label="Test prop text" danger />)
     const button = container.firstChild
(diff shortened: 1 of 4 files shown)


burbuli8ra (commented):
Same tests are failing!
Please lets fix them before merging this PR! 😄

jacekkolasa (changes requested):
Sorry @christophidesp but those colors are used in dashboard! We would need to fix it in there, first.
(Also perhaps in cloud-iframes, but i'm not sure, need to check). I can help you with it

christophidesp:
@jacekkolasa they were already declared some lines above, so we had duplicate declarations for error and warning.

jacekkolasa (approved):
@christophidesp didn't noticed it! Thanks :)

PR #381, opened 2022-10-10
chore(drop): update declaration file for Drop component

burbuli8ra:
#### Solution
Update declaration file of `Drop` component, so that `cloud-frontend` does not give TS warnings!


AlexNti (approved):

src/components/drops/drop/index.d.ts
```diff
@@ -8,8 +8,8 @@ export interface DropProps extends FlexProps, AlignProps, StretchProps {
   children: any
   canHideTarget?: boolean
   [key: string]: any
-  backdrop: boolean
-  backdropProps: any
+  backdrop?: boolean
+  backdropProps?: any
```
jacekkolasa, line 12: +1

PR #388, opened 2022-10-19
feat(input): add input range component

burbuli8ra:
This PR adds the `InputRange` component on the library


src/components/input/range/styled.js
```diff
@@ -0,0 +1,25 @@
+import styled from "styled-components"
+
+export const InputRange = styled.input.attrs({ type: "range" })`
+  -webkit-appearance: none;
+  height: 2px;
+  background: #00ab4440;
```
AlexNti, line 6: I suggest to use the colors from our color file or using the getColor function

src/components/input/range/index.js
```diff
@@ -0,0 +1,21 @@
...
+const InputRange = ({ max = 100, min = 0, step = 1, value = 0, ...rest }) => {
+  const getBackgroundSize = () => ({ backgroundSize: `${(value * 100) / max}% 100%` })
+
+  return (
+    <StyledInputRange
+      data-testid="rangeInput"
+      max={max}
+      min={min}
+      step={step}
+      style={getBackgroundSize()}
```
AlexNti, line 13: We can move `getBackgroundSize` inside the InputRange function and get rid of the `style` property.

src/components/input/range/index.js
```diff
@@ -0,0 +1,21 @@
+import React, { useState } from "react"
```
AlexNti, line 1: Does this useState is being used ?

src/components/input/range/index.d.ts
```diff
@@ -0,0 +1,19 @@
...
+export interface RangeInputProps {
+  "data-testid"?: string
+  max?: number
+  min?: number
+  onChange: (e: ChangeEvent<HTMLInputElement>) => void
+  onClick?: (e: SyntheticEvent<HTMLButtonElement>) => void
+  onInput?: (e: SyntheticEvent<HTMLButtonElement>) => void
+  step?: number
+  value: number
+  [key: string]: any
```
AlexNti, line 12: What do what you like to achieve here? By adding this property it means that whatever property you pass to the `InputRange` it will be valid, e.x if i pass the property `randomPorperty = "adads"` typescript will not warn me that i have passed a property that does not exist in the types.

src/components/input/range/index.d.ts
```diff
@@ -0,0 +1,19 @@
+import { ChangeEvent, FC, SyntheticEvent } from "react"
+
+export interface RangeInputProps {
+  "data-testid"?: string
+  max?: number
+  min?: number
+  onChange: (e: ChangeEvent<HTMLInputElement>) => void
+  onClick?: (e: SyntheticEvent<HTMLButtonElement>) => void
+  onInput?: (e: SyntheticEvent<HTMLButtonElement>) => void
```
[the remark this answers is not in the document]
burbuli8ra: Since we are implementing this component to replace the previous implementation I would have to say that it would be good to give the users the option to have the ones that we know at least that we are going to use! 😄

src/components/input/range/index.d.ts
burbuli8ra: It's a pattern that all components of our library have!
That's why I have also included it here! 😄

AlexNti (approved):

PR #391, opened 2022-10-26
fix no-key warnings

jacekkolasa:
fix 3 no-key warnings when no actions are provided to the table

### DIFF
--- a/src/components/tableV2/netdataTable.js
+++ b/src/components/tableV2/netdataTable.js
@@ -4,11 +4,11 @@ import React, { useEffect, useMemo, useRef, useState } from "react"
 import Table, { Pagination } from "./base-table"
 
 import {
+  flexRender,
   getCoreRowModel,
   getFilteredRowModel,
   getPaginationRowModel,
   getSortedRowModel,
-  flexRender,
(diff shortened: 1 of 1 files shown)


src/components/tableV2/netdataTable.js
```diff
@@ -661,11 +662,10 @@ const renderActionWithDropdown = ({
       const actionRef = useRef()
 
       return (
-        <>
+        <React.Fragment key={id}>
```
AlexNti, line 665: I think its safe to complete remove the Fragment here.

AlexNti (approved):

src/components/tableV2/netdataTable.js
jacekkolasa: @AlexNti There are 2 components inside - Action and ColumnsMenu, perhaps you missed it because of how GH collapses the code?

burbuli8ra (approved):

src/components/tableV2/netdataTable.js
AlexNti: Oh, yes you correct. I missed that

PR #400, opened 2022-11-14
Feat/column pinning

AlexNti:


burbuli8ra (approved):
On a general note, I am not quite sure about defining a fixed height on the columns!
I understand that we will need to "protect" our sticky table from breaking if the height of lines on the 2 tables are not equal, but maybe we could only set a fixed height when the table has the prop `enableColumnPinning` set to `true`

PR #409, opened 2022-11-21
chore(netdata table): add border to last sticky column cells

burbuli8ra:
This PR implements the missing border on cells of the last column on `sticky` columns table

### DIFF
--- a/src/components/tableV2/core/base-table.js
+++ b/src/components/tableV2/core/base-table.js
@@ -119,8 +119,6 @@ Table.HeadCell = forwardRef(
       children,
       align = "left",
       width,
-      maxWidth,
-      minWidth,
       tooltipText,
       filter,
       onMouseDown,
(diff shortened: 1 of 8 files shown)


src/components/tableV2/core/base-table.js
```diff
@@ -255,39 +253,41 @@ Table.Body = forwardRef(({ children, ...props }, ref) => (
...
+        onClick={handleClick}
         padding={[3]}
+        ref={ref}
         sx={{
           textAlign: align,
           height: "65px",
           maxHeight: "65px",
           whiteSpace: "nowrap",
           ...cellStyles,
+          ...(pinnedStyles || {}),
```
jacekkolasa, line 286: consistent with present approach would be to have `pinnedStyles = {}` when destructuring (just as `styles`)

jacekkolasa (approved):

PR #410, opened 2022-11-21
allow pseudos to receive multiple styles

AlexNti:
Pseudos now can be used with any style that we like to pass e.x. 
```
   _before={{
        content: '":"',
        position: "absolute",
        top: "0",
        bottom: "0",
        display: "flex",
        alignItems: "center",
        left: "-8px",
        width: "8px",
      }}
```

### DIFF
--- a/src/components/templates/mixins/pseudos.js
+++ b/src/components/templates/mixins/pseudos.js
@@ -83,8 +83,6 @@ export default ({ theme, ...props }) => {
       ${pseudoSelectors[pseudoProp]}{ 
         ${styles} 
       }`
-
-      if (props) console.log(pseudo)
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/components/templates/mixins/pseudos.js
```diff
@@ -50,19 +75,16 @@ export default ({ theme, ...props }) => {
...
+      const styles = calculateStyles({ theme, ...pseudoStyles })
 
       pseudo =
         pseudo +
         `
       ${pseudoSelectors[pseudoProp]}{ 
         ${styles} 
       }`
+
+      if (props) console.log(pseudo)
```
burbuli8ra, line 87: Do we need this?

jacekkolasa (approved):

PR #411, opened 2022-11-23
Feat/virtualize netdata table

AlexNti:
By default table is virtualized and allow table to be manual paginated in the scenario of `enablepagination===false`.


novykh (approved):
It's good to go, we need to revisit static height for rows.

PR #412, opened 2022-11-24
Feat/infinity scroll

AlexNti:
Allow NetdataTable to use the infiniteScroll functionallity.

### DIFF
--- a/src/components/tableV2/core/rows.js
+++ b/src/components/tableV2/core/rows.js
@@ -19,6 +19,7 @@ const Rows = ({
   testPrefix,
   testPrefixCallback,
   scrollParentRef,
+  overscan,
 }) => {
   const { state, updateState } = useSharedTable({})
   const { rows } = table.getRowModel()
@@ -27,7 +28,7 @@ const Rows = ({
     count: rows.length,
     getScrollElement: () => scrollParentRef.current,
     estimateSize: () => CELL_HEIGHT,
-    overscan: 30,
+    overscan: overscan || 30,
   })
 
   const virtualRows = virtualizer.getVirtualItems()
--- a/src/components/tableV2/core/rows.js
+++ b/src/components/tableV2/core/rows.js
@@ -27,7 +27,7 @@ const Rows = ({
     count: rows.length,
     getScrollElement: () => scrollParentRef.current,
     estimateSize: () => CELL_HEIGHT,
-    overscan: 20,
+    overscan: 30,
   })
 
   const virtualRows = virtualizer.getVirtualItems()
--- a/src/components/tableV2/hooks/useInfiniteScroll.js
+++ b/src/components/tableV2/hooks/useInfiniteScroll.js
@@ -1,9 +1,14 @@
 import { useEffect, useMemo, useState } from "react"
 
-const useIniniteScroll = (
+const useIniniteScroll = ({
   service,
-  { target, onError, onSuccess, onDataend, updateTableData, threshold = 100 }
-) => {
+  target,
+  onError,
+  onSuccess,
+  onDataend,
+  updateTableData,
+  threshold = 100,
+}) => {
   const [lastReceivedData, setLastReceivedData] = useState()
   const [loading, setLoading] = useState(false)
 
--- a/src/components/tableV2/hooks/useInfiniteScroll.js
+++ b/src/components/tableV2/hooks/useInfiniteScroll.js
@@ -1,0 +1,113 @@
+import { useEffect, useMemo, useState } from "react"
+
+const fakeData = () => [
+  {
+    nodes: "node51",
+    alerts: 15,
+    user: "mitsos",
+    disabled: true,
+    status: "stale",
+    untouchable: "true",
+  },
+  {
+    nodes: "node52",
+    alerts: 15,
+    user: "mitsos",
+    disabled: true,
+    status: "stale",
+    untouchable: "true",
+  },
+  {
+    nodes: "node53",
+    alerts: 15,
+    user: "mitsos",
+    disabled: true,
+    status: "stale",
+    untouchable: "true",
+  },
+  {
+    nodes: "node54",
+    alerts: 15,
+    user: "mitsos",
+    disabled: true,
+    status: "stale",
+    untouchable: "true",
+  },
+  {
+    nodes: "node55",
+    alerts: 15,
+    user: "mitsos",
+    disabled: true,
+    status: "stale",
+    untouchable: "true",
+  },
+  {
+    nodes: "node56",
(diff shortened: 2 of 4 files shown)


burbuli8ra (approved):

src/components/tableV2/hooks/useInfiniteScroll.js
```diff
@@ -0,0 +1,67 @@
...
+    const scrollTop = el.scrollTop
+    const scrollHeight = el.scrollHeight
+    const clientHeight = el.clientHeight
+    if (scrollHeight - scrollTop <= clientHeight + threshold) {
+      loadMore()
+    }
+  }
+  useEffect(() => {
+    if (!target?.current || loading || !service) return
+    const lisener = () => {
```
novykh, line 53: typo

src/components/tableV2/hooks/useInfiniteScroll.js
```diff
@@ -0,0 +1,67 @@
...
+      setLastReceivedData(currentData)
+      onSuccess?.(currentData)
+      setLoading(false)
+    } catch (err) {
+      onError?.(err)
+      setLoading(false)
+    }
+  }
+
+  const loadMore = () => {
```
novykh, line 34: Both this, `scrollMethod ` and `runService` should be created inside the useEffect, no need to recreate on every rerender.

src/components/tableV2/hooks/useInfiniteScroll.js
```diff
@@ -0,0 +1,67 @@
...
+  onDataend,
+  updateTableData,
+  threshold = 100,
+}) => {
+  const [lastReceivedData, setLastReceivedData] = useState()
+  const [loading, setLoading] = useState(false)
+
+  const hasDataEnded = useMemo(() => {
+    if (!onDataend) return false
+    return onDataend(lastReceivedData)
```
novykh, line 17: typo

src/components/tableV2/hooks/useInfiniteScroll.js
```diff
@@ -0,0 +1,67 @@
...
+  onError,
+  onSuccess,
+  onDataend,
+  updateTableData,
+  threshold = 100,
+}) => {
+  const [lastReceivedData, setLastReceivedData] = useState()
+  const [loading, setLoading] = useState(false)
+
+  const hasDataEnded = useMemo(() => {
```
novykh, line 15: this should be passed as a dependency in the `useEffect`

src/components/tableV2/hooks/useInfiniteScroll.js
```diff
@@ -0,0 +1,67 @@
...
+    if (!onDataend) return false
+    return onDataend(lastReceivedData)
+  }, [lastReceivedData])
+
+  const runService = async () => {
+    try {
+      const currentData = await service(lastReceivedData)
```
novykh, line 22: This is in order to get the page/cursor payload? In any case `lastReceivedData` should be passed as a dependency to the `useEffect`

AlexNti: You are right, i made a fix that introduces a new hook `useScrollListener` this will accept the `target` and `event handler` as input. The event handler will be saved as `ref` to not trigger a rerender.

src/components/tableV2/core/rows.js
```diff
@@ -27,7 +27,7 @@ const Rows = ({
     count: rows.length,
     getScrollElement: () => scrollParentRef.current,
     estimateSize: () => CELL_HEIGHT,
-    overscan: 20,
+    overscan: 30,
```
novykh, line 30: nit: This can be a prop

src/components/tableV2/hooks/useInfiniteScroll.js
```diff
@@ -0,0 +1,139 @@
...
+//       })
+//     }, 1000)
+//   })
+// }
+
+const useScrollListener = (target, handler) => {
+  const listenerRef = useRef(handler)
+  listenerRef.current = handler
+
+  const listener = () => listenerRef.current()
```
novykh, line 71: nit: This can be created inside the useEffect, to avoid recreating in each rerender.

src/components/tableV2/hooks/useInfiniteScroll.js
```diff
@@ -0,0 +1,139 @@
...
+
+  useEffect(() => {
+    if (!target?.current) return
+    target.current.addEventListener("scroll", listener)
+
+    return () => {
+      if (!target.current) return
```
novykh, line 78: Not needed since this will never happen, since line 74 will not create this cleanup function.

src/components/tableV2/hooks/useInfiniteScroll.js
```diff
@@ -0,0 +1,139 @@
...
+    }
+    const scrollTop = el.scrollTop
+    const scrollHeight = el.scrollHeight
+    const clientHeight = el.clientHeight
+    if (scrollHeight - scrollTop <= clientHeight + threshold) {
+      loadMore()
+    }
+  }
+
+  useScrollListener(target, () => {
```
novykh, line 133: All the methods above will be created and recreated in every rerender even if you don't have `service` .

src/components/tableV2/hooks/useInfiniteScroll.js
AlexNti: Sometime on refreshing the page on stories, i got that target is undefined thats why i put it there.

src/components/tableV2/hooks/useInfiniteScroll.js
AlexNti: The Listener is inside a useRef,this means that useEffect will not be affected by it, it will not trigger a rerender.

src/components/tableV2/hooks/useInfiniteScroll.js
novykh: Target is a ref? Now I saw the `.current` why is this a ref? If it's a ref, then deps `[target]` doesn't do anything, since target itself doesn't change.
OK, so leave the `if` there, but remove the dependency from the array.

src/components/tableV2/hooks/useInfiniteScroll.js
novykh: The rerender will come from the parent. Put a console.log on line 132 to see that the above functions will be recreated everytime, even though you are not using them when there is no service.

AlexNti: You right , it triggers a rerender but i this happening because i have a logic error. Should be `!service` instead of `service` on the if statement.

PR #413, opened 2022-11-29
remove redundant height from labels and add margin bottom

christophidesp:
Before
Screenshot 2022-11-29 at 11 12 26
After
Screenshot 2022-11-29 at 11 12 58

### DIFF
--- a/src/components/input/styled.js
+++ b/src/components/input/styled.js
@@ -50,17 +50,16 @@ export const StyledLabel = styled.label`
   display: block;
   ${disabledCursorSupport};
 `
-export const LabelRow = styled.div`
-  width: 100%;
-  flex-shrink: 0;
-  flex-grow: 0;
+export const LabelRow = styled(Flex).attrs({
+  alignItems: "center",
+  color: "text",
+  flex: false,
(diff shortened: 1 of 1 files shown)


src/components/input/styled.js
```diff
@@ -52,7 +52,6 @@ export const StyledLabel = styled.label`
 `
 export const LabelRow = styled.div`
```
burbuli8ra, line 53: This comment is not related to your change, but I think that when we make an update to a component that seems to be outdated we could get it up-to-date and in this case I think we could replace `div` with `Flex`.
I am adding a suggestion bellow, if you want you can merge it as is! 😄 

```suggestion
export const LabelRow = styled(Flex).attrs({
  alignItems: "center",
  color: "text",
  flex: false,
  width: "100%",
})`
  font-size: 14px;
  font-style: normal;
  font-weight: bold;
  line-height: 18px;
`
```

burbuli8ra (approved):

src/components/input/styled.js
christophidesp: Updated :)

PR #426, opened 2023-01-09
fix(dropdown): make search on label value optional

burbuli8ra:
This PR fixes issue [Cloud Frontend #695](https://github.com/netdata/netdata-cloud/issues/695)

#### Solution
* Check if type of label is "string" before applying `toLowerCase` method
* Update **Run Tests** to run on every push, not only when pushed on `master` branch

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,7 +1,10 @@
 name: tests
 
 on:
+  pull_request:
   push:
+    branches:
+      - master
   workflow_dispatch:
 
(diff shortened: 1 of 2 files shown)


jacekkolasa (approved):

novykh (commented):
Why? This step runs the tests.

burbuli8ra:
@novykh I was just making some checks!
I removed the rule to run when pushed on `master` finally!
Let me know if it's ok!

PR #431, opened 2023-01-17
fix(notifications): UI updates for notification integrations feature

burbuli8ra:
This PR includes the following updates:
* **Input:** Update font size for `tiny` size
* **Theme:** Update color of `placeholder` value

### DIFF
--- a/src/components/input/styled.js
+++ b/src/components/input/styled.js
@@ -29,14 +29,14 @@ export const StyledInput = styled.input.attrs({ round: true })`
   flex-grow: 0;
   ${({ iconLeft }) => iconLeft && "padding-left: 0"};
   ${({ iconRight }) => iconRight && "padding-right: 0"};
-  font-size: 14px;
+  font-size: ${({ size }) => (size === "tiny" ? "12px" : "14px")};
   line-height: 18px;
   color: ${({ disabled }) => (disabled ? getColor("placeholder") : getColor("textDescription"))};
   background: ${({ disabled }) =>
     disabled ? getColor("mainBackgroundDisabled") : getColor("mainBackground")};
 
   &::placeholder {
-    font-size: 14px;
+    font-size: ${({ size }) => (size === "tiny" ? "12px" : "14px")};
     line-height: 18px;
     color: ${getColor("placeholder")};
     opacity: 1;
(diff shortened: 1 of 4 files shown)


src/components/input/styled.js
```diff
@@ -58,7 +58,7 @@ export const LabelRow = styled(Flex).attrs({
 })`
   font-style: normal;
   font-weight: bold;
-  font-size: 14px;
+  font-size: ${({ size }) => (size === "tiny" ? "12px" : "14px")};
```
jacekkolasa, line 61: @burbuli8ra how about `small` keyword, since in `typography.js` we use it for `12px`? (not a strong opinion, i can understand this flagName is more about the input, not just the text)

src/theme/default/colors.js
```diff
@@ -48,7 +48,7 @@ const appColors = {
   sectionHeaderBackground: rawColors.neutral.limedSpruce,
   sectionTitle: rawColors.neutral.bluebayoux,
   sectionDescription: rawColors.neutral.regentgrey,
-  placeholder: rawColors.neutral.iron,
+  placeholder: rawColors.neutral.regentgrey,
```
jacekkolasa, line 51: we'll need to check if it looks ok everywhere it is used in cloud-fe, or did you already?

jacekkolasa (approved):

src/components/input/styled.js
burbuli8ra: Actually @jacekkolasa you have a valid point here, but `size` was not introduced by me... 😄 
I have used the already existing property (used for example [here](https://github.com/netdata/netdata-ui/blob/master/src/components/input/use-input-styles.js)) and have just updated the missing styles to align with what is on the design system.
As already discussed with @christophidesp we need at some point to review in general the inputs/etc. to align everything and cleanup the code as well!
For the time, I am just going to keep this update instead and discuss a potential refactor when we do the general alignment needed for those components! 😄

src/theme/default/colors.js
burbuli8ra: Yes, indeed!
@christophidesp has already commited a changes needed for this update on the relevant PR I have opened on `cloud-frontend`! We know that we need to keep an eye if anything looks wired with this update and do the proper updates on `cloud-frontend`! 😄

PR #437, opened 2023-02-08
chore(netdata table): add support for custom tooltip component

burbuli8ra:


jacekkolasa (approved):

src/components/tableV2/components/action.js
```diff
@@ -58,6 +59,7 @@ const Action = forwardRef(
     }
 
     const Component = label ? Button : IconButton
+    const TooltipComponent = CustomTooltip || Tooltip
```
novykh, line 62: use the default value on top to pass tooltip if `undefined`

novykh (approved):

PR #440, opened 2023-02-17
data-ga tagging updates

jacekkolasa:


src/components/tableV2/features/globalControls.js
```diff
@@ -11,13 +11,14 @@ const GlobalControls = ({
...
 }) => {
+  const wordsCount = searchValue?.split(" ").filter(x => !!x).length
   return (
     <Flex width="100%" zIndex={10} background="mainBackground" padding={[0, 0, 4]}>
       {handleSearch && (
         <Flex width={{ max: 100, base: "40%" }}>
           <SearchInput
             data-testid="table-global-search-filter"
-            data-ga={`${dataGa}::search::table-filter`}
+            data-ga={`${dataGa}::search-${wordsCount}-words::table-filter`}
```
burbuli8ra, line 21: Why is this needed here?

jacekkolasa: @burbuli8ra it's addressing [this requirement](https://github.com/netdata/product/issues/3064#user-content-functions-low-prio)
```
 Search bar interactions - how many users use this search bar? ideally capturing the words used
```

burbuli8ra: I mean this will constantly be updated while user types, so I cannot really understand how it is helpful for analytics

jacekkolasa: @burbuli8ra i'm not sure which tools we do have at our disposal in PostHog or GA, i thought that when it will be deployed, we can experiment and modify it if it won't be useful.

burbuli8ra (approved):

PR #449, opened 2023-03-08
fix table title, set new color vars, for dropdown table, progress bars etc

christophidesp:


src/components/tableV2/core/base-table.js
```diff
@@ -27,7 +27,7 @@ const HeaderCell = styled(Box)`
 
 const Table = forwardRef(({ children, width, ...props }, ref) => (
   <Box
-    sx={{ borderCollapse: "separate", "table-layout": "fixed" }}
+    sx={{ borderCollapse: "separate", "table-layout": "fixed", borderSpacing: "0px" }}
```
burbuli8ra, line 30: `0` is not working?

burbuli8ra (approved):

src/components/tableV2/core/base-table.js
christophidesp: it does :)

PR #450, opened 2023-03-08
increase pinned table zindex to prevent thead overlapping

christophidesp:


src/components/tableV2/features/columnPinning.js
```diff
@@ -25,7 +25,7 @@ const ColumnPinning = ({
       sx={{
         position: "sticky",
         left: 0,
-        zIndex: 2,
+        zIndex: 11,
```
burbuli8ra, line 28: Is it needed to be that large?

christophidesp: Yes. It is only 11 after all, it is not like it is 9999999 :P

burbuli8ra: Well 3 from 11 has a difference... that's why I asked... if t is possible to keep it bellow 11 it would be great, if not then ok! :)

burbuli8ra (approved):

PR #454, opened 2023-03-27
Confirmation dialog - Disable decline button on confirmation loading state

kapantzak:


src/components/confirmation-dialog/confirmation-dialog.js
```diff
@@ -26,6 +26,7 @@ const ConfirmationDialog = ({
   title,
   showConfirmLoading,
   disableConfirmOnLoading,
+  disableDeclineOnLoading = true,
```
burbuli8ra, line 29: I am leaving this comment here, for 2 reasons...
1. I am not a freak of the positive labeled variables, but when possible I prefer of having them named like that.
2. In general, I am also not a fan of having boolean component props declared and initialised with a `true` value, if this is not a necessity, because this means that whenever the component needs to be used the user will have to write this 👇 
```
<ConfirmationDialog disableDeclineOnLoading={false} {...etc} />
```

instead of this 👇 
```
<ConfirmationDialog disableDeclineOnLoading {...etc} />
```

For all the above, I would recommend to use `enableDeclineOnLoading` instead and either set it's initial value to `false` or not (it's up to you) and just pass it bellow as this 👇 
```
disabled={loading && !enableDeclineOnLoading}
```

kapantzak: Agree with you regarding the positive labeled variables, but in this case, I think that renaming this to `enableDeclineOnLoading` implies that the `Decline` button in always disabled unless we explicitly tell it to be enabled on loading.
Furthermore, there is also the `disableConfirmOnLoading` prop name that does the same think for the confirm button.

burbuli8ra: Regarding this

> I think that renaming this to enableDeclineOnLoading implies that the Decline button in always disabled unless we explicitly tell it to be enabled on loading.
Shouldn't it be the normal case?

I also, might have missed the `disableConfirmOnLoading` back then... not sure if I was one of the approvers! 😄

burbuli8ra (approved):

PR #464, opened 2023-05-03
fix(netdata table): remove useCallback from onSorting

burbuli8ra:
This PR removes `useCallback` from sorting handler of table, in order for the user to be able to regenerate the function passed when needed attribute values have been altered


novykh (commented):
Better to pass the dependency needed.

novykh (approved):

PR #466, opened 2023-05-20
Add right column pinning to table

kapantzak:
This PR adds the ability to define right-pinned columns in NetdataTable component.

```
const columnPinning = { right: ["actions"] }
```

It also handles the rows heights and syncs the pinned columns height with the respective table row (if they differ)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/netdata-ui",
-  "version": "2.16.6",
+  "version": "2.16.7",
   "description": "netdata UI kit",
   "main": "./lib/index.js",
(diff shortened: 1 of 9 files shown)


burbuli8ra:
@kapantzak do we need the table to support right pinning?
I am asking because the table is build to support only what will be used, so if for some reason we need the right column pinning then ok, if we don't... I would suggest not to add it! 😄

christophidesp:
@burbuli8ra yes we will need it. We need to be able to pin actions column on the last column of the table.

burbuli8ra (approved):

PR #474, opened 2023-06-16
Update modals design

kapantzak:
Update the design of modal according to the [new design system](https://www.figma.com/file/qvOZOmj058lhINHx4vowxC/Assistant?type=design&node-id=19-8507&t=eAoGnW3O7Ex8RQu6-0)

Screenshot 2023-06-16 at 4 09 41 PM

### DIFF
--- a/src/components/confirmation-dialog/styled.js
+++ b/src/components/confirmation-dialog/styled.js
@@ -57,7 +57,7 @@ export const Dialog = styled(Modal).attrs({
 export const Header = styled(ModalHeader).attrs({
   alignItems: "center",
   padding: [0],
-  background: "none",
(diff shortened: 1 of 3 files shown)


src/components/confirmation-dialog/styled.js
```diff
@@ -57,6 +57,7 @@ export const Dialog = styled(Modal).attrs({
 export const Header = styled(ModalHeader).attrs({
   alignItems: "center",
   padding: [0],
+  background: "none",
```
novykh, line 60: Maybe you mean to put an empty value here? check the background mixin.

novykh (approved):

PR #477, opened 2023-06-30
Add container styles to input

kapantzak:
- ~Add width mixin~
- Move reusable helper methods (`getDimensions` and `getDimension`) out of margin mixin


kapantzak:
I can revert the `Move reusable helper methods out of margin mixin` commit if don't want this change

novykh (approved):

PR #494, opened 2023-10-24
Redesign

novykh:


christophidesp:
Not sure what has being redesigned
Are there specific items to look at?

From a quick view I saw that inputs have a green bg in dark theme, which is not expected
Screenshot 2023-10-24 at 17 27 12

novykh:
> Not sure what has being redesigned Are there specific items to look at?
> 
> From a quick view I saw that inputs have a green bg in dark theme, which is not expected Screenshot 2023-10-24 at 17 27 12

ah good catch, this was pushed by mistake. But now that you mentioned it, we need a way to know for search inputs, when they have a value in them. To be more prominent, so users can see that they have a filter set in them.

christophidesp:
@novykh here is the link for inputs styleguide
https://www.figma.com/file/v0QSRTMqaUQ7OOxL6UP151/Design-System?type=design&node-id=51%3A396&mode=design&t=vSPEPxFLG5rFWHGE-1

The states documented there are:
Idle
Hover
Active
Hover when already Active
ReadOnly

I believe that they cover the need of showing if something has already been selected in Active state of input.

Besides that, what else should I check for changes?

kapantzak (approved):

novykh:
I think pills, yeah it's a little vague, but since I did some ui changes based on your latest, I called it redesign 😝

PR #496, opened 2023-11-03
Add custom content to table header

kapantzak:
Adds custom content to header

Screenshot 2023-11-03 at 11 23 18 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/netdata-ui",
-  "version": "4.1.15",
+  "version": "4.1.16",
   "description": "netdata UI kit",
   "main": "dist/index.js",
   "module": "dist/es6/index.js",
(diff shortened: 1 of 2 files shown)


novykh (commented):
Let's close it for now, since we have `title` prop, that you can render whatever you need in there. The only thing is it will be positioned before the search input. If there is a really big reason to put it there, then pass the element as prop and not the Component. 
```
headerChildren={<Flex>Something</Flex>}
```

kapantzak:
> Let's close it for now, since we have `title` prop, that you can render whatever you need in there. The only thing is it will be positioned before the search input. If there is a really big reason to put it there, then pass the element as prop and not the Component.
> 
> ```
> headerChildren={<Flex>Something</Flex>}
> ```

I don't see any `headerChildren` used anywhere in the ui repo, do I miss something?

And placing it as a title, feels a bit strange to me:

Screenshot 2023-11-06 at 11 06 28 PM

novykh (approved):

PR #499, opened 2023-11-14
Deactivate transitions in toggle

kapantzak:


src/components/toggle/styled.js
```diff
@@ -25,7 +25,7 @@ export const StyledToggle = styled.div`
     props.disabled ? getColor("mainBackgroundDisabled") : getColor("mainBackground")};
   border: 1px solid ${getColor("border")};
   border-radius: 100px;
-  transition: all 150ms;
+  transition: ${({ noTransition }) => (noTransition ? "unset" : "all 150ms")};
```
novykh, line 28: Maybe keep it without transitions as default?

novykh (approved):

src/components/toggle/styled.js
kapantzak: Agree, changed prop to `withTransition`

PR #550, opened 2025-01-31
Fix memoized table object

kapantzak:
When we pass a value to `dataColumns` prop of the `Table` component, that changes over re-renders, `BodyHeader` component cannot distinguish between different versions of `table` object and keeps the first version of it.

~By adding `timestamp` prop to `table` object each time we create it, we ensure that `memo` recognizes each version and re-renders `BodyHeader` component to reflect latest state.~

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/netdata-ui",
-  "version": "4.8.14",
+  "version": "4.8.15",
   "description": "netdata UI kit",
   "main": "dist/index.js",
   "module": "dist/es6/index.js",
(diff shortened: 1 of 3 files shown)


novykh (changes requested):
This has a number of performance issues. Better to update the selector on the component you want to rerender when columns change.
I think you need this: https://github.com/netdata/netdata-ui/blob/6444b54876f50419c24a14c153548531280f5272/src/components/table/body/header/index.js#L13

But again I would proceed with caution and checking the performance implications - especially on tables that do not need this check.

PR #621, opened 2026-05-13
Defer measureElement

kapantzak:


kapantzak:
Doesn't work properly

PR #631, opened 2026-07-13
Address large data table review findings

novykh:


ktsaou (approved):

src/components/table/largeDataSelection.js
```diff
@@ -1,35 +1,59 @@
...
 
 export const getIsSomeRowsSelected = (source, rowSelection) => {
-  const selectedCount = Object.keys(rowSelection).length
-  return selectedCount > 0 && selectedCount < source.getFlatRowCount()
+  let selectedCount = Object.keys(rowSelection).length
+  if (!selectedCount) return false
+
+  let selectableCount = 0
+  let selectedSelectableCount = 0
+  source.forEachExportRow((row, id) => {
```
ktsaou, line 34: `getIsSomeRowsSelected()` is called directly while rendering the selection header (`useColumns/useRowSelection.js:13`), but this loop now traverses every export row. With 50,000 rows and one selected row, a local 1,000-call benchmark changed from less than 0.001 ms/call on the merged baseline to 5.89 ms/call here (about 5.9 seconds total). This makes unrelated renders scale with the entire room and recreates the stalls this work is intended to remove. Please precompute/maintain the selectable-row count and an ID/selectability lookup with the source, then derive the partial-selection state from selected IDs only; the render path must not traverse the full source.

src/components/table/body/header/cell.js
```diff
@@ -122,7 +123,7 @@ const BodyHeaderCell = ({
             <Label
               as={column.columnDef.labelAs}
               {...column.columnDef.labelProps}
-              data-table-header-tooltip={headerTooltipContent}
+              {...{ [overflowTooltipAttribute]: headerTooltipContent }}
```
ktsaou: This removes the dedicated `data-table-header-tooltip` ownership marker and replaces it with the generic body-cell marker. The pending cloud integration deliberately delegates header tooltips through `[data-table-header-tooltip], th` and checks truncated descendants; most tables do not pass the local `overflowTooltip` controller, so their header tooltips will stop working. Expanding the global selector to the generic marker is also unsafe because nodes/alerts opt into a local controller and would then have two controllers owning the same target. Please preserve the distinct header marker and reserve `data-overflow-tooltip` for the opt-in local/body-cell controller. The current test forces `isOverflowing: () => true`, so it does not exercise this real ownership/integration path.

ktsaou (changes requested):
Two blocking regressions remain on the current head: selection-state rendering scales with the full data set, and the header-specific tooltip ownership contract is replaced by the generic body-cell contract. Both undermine the large-table work and need to be addressed before merge. The sorting-cache trade-off discussed separately is non-blocking.

src/components/table/largeDataSelection.js
novykh: Full header-state benchmark, 50,000 rows and one selection:
PR 629: 7.46 ms
PR 631: 7.89 ms
The reported isolated getIsSomeRowsSelected() regression is real, but measuring it alone hides the pre-existing scan in “all.” An indexed source would still be a worthwhile optimization, but this is inherited performance debt, not a new one-third-frame regression. Reverting to the old formula would also break disabled-row semantics.

ktsaou: Agreed. The isolated helper benchmark overstated the full header-path regression because the merged baseline already traversed the complete source in getIsAllRowsSelected(). Reproducing the complete calculation confirms only a small delta for sparse selection; dense selections remain slower, but the missing index is inherited performance debt and reverting would break the corrected disabled-row semantics. This is not a blocker for this PR.

src/components/table/body/header/cell.js
ktsaou: Fixed in 73845887: the dedicated data-table-header-tooltip marker is restored and the test now protects the ownership boundary. Thank you.

ktsaou (approved):
The header-tooltip ownership regression is fixed. The selection discussion is resolved: the complete render path already carried the source traversal, while this change preserves correct disabled-row semantics. Focused table tests pass; approved pending the required CI check.

PR #637, opened 2026-08-07
v5.5.7

car12o:
Adds the `group_email_colored` SVG asset under `src/components/icon/assets/integrations/` and registers it as `integrationGroupEmailColored` in `iconsList.js` so it is available through the `Icon` component.


src/components/icon/assets/integrations/group_email_colored.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
novykh, line 1: is this tested n the browser in both themes? I think it has too many inline color styling.

car12o: No, I don't know how I can deploy this. Just craft a simple HTML page to render in the browser, and it looks good :smile: 
Can you take it from here pls?

novykh: Yep that's why I asked, I will test it and merge it thanks man

car12o: Thanks man :bow:

kapantzak (approved):

